Issue #73, opened 2017-10-09
Make Timer use in hub related operations non-reentrant


timlaverty:
# Type of issue

- [x ] Bug
- [ ] New feature
- [ ] Enhancement

# Description
Change the Timer use so operations that involve touching the hub aren’t reentrant; i.e. add a Pause and Unpause method to the Timer and implement it in each of SendTelemetry and UpdateDesiredProperties....

dluc:
done

Issue #71, opened 2017-10-09
Change device registration in to register devices in batches


timlaverty:
# Type of issue

- [x ] Bug
- [ ] New feature
- [ ] Enhancement

# Description
 Change device registration in simulation to register/create devices in batches.  This should be configurable but by default should register <10 at a time (the S1 limitation). ...

PR #78, opened 2017-10-12
Limit registry and twin operations as per configuration

dluc:
# Types of changes 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

Allow to configure at what speed the simulation can use the device registry and the device twins, to comply with the service size and limits, e.g. how many connections per second, how many twin reads/writes per second. This change covers https://github.com/Azure/device-simulation-dotnet/issues/71 and needs to be complemented by https://github.com/Azure/device-simulation-dotnet/issues/73

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/78)


Services.Test/Concurrency/PerMinuteCounterTest.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public PerMinuteCounterTest(ITestOutputHelper logger)
+        {
+            log = logger;
+            this.targetLogger = new TargetLogger(logger);
+        }
+
+        /**
+         * Calls are slowed down only *after* reaching the limit for events
+         * per minute. So, when the limit is 60 events/minute, 60 events should
```
timlaverty: do limitations/minute make sense to test?  per second and per day are the hub throttles.

Services.Test/Concurrency/PerSecondCounterTest.cs
```diff
@@ -0,0 +1,246 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Concurrency;
+using Services.Test.helpers;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Services.Test.Concurrency
```
timlaverty: nit: Why the concurrency namespace?  ThrottlingLimits = better name?

Services.Test/Concurrency/PerSecondCounterTest.cs
```diff
@@ -0,0 +1,246 @@
...
+        public void ItWorksWhenNoThrottlingIsNeeded()
+        {
+            // Arrange
+            var target = new PerSecondCounter(10, "test", this.targetLogger);
+
+            // Act
+            var paused = false;
+            for (var i = 0; i < 10; i++)
+            {
+                paused = paused || target.IncreaseAsync().Result;
```
timlaverty: nit: could check for !paused and break earlier if you wanted to make the test return quicker.

Services/Concurrency/RateLimiting.cs
```diff
@@ -0,0 +1,124 @@
...
+        Task<T> LimitMessagesAsync<T>(Func<Task<T>> func);
+        Task LimitMessagesAsync(Func<Task> func);
+    }
+
+    public class RateLimiting : IRateLimiting
+    {
+        // Use separate objects to reduce internal contentions in the lock statement
+
+        private readonly PerSecondCounter connections;
+        private readonly PerMinuteCounter registryOperations;
```
timlaverty: what is limited per minute?

Services/Concurrency/RateLimiting.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            this.twinWrites = new PerSecondCounter(
+                config.TwinWritesPerSecond, "Twin writes", log);
+
+            this.messaging = new PerDayCounter(
+                config.MessagesPerDay, "Messages", log);
+
+            // The class should be a singleton, if this appears more than once
+            // something is not setup correctly and the rating won't work.
+            // TODO: enforce the single instance, compatibly with the use of
```
timlaverty: github issue to track?

Services/Concurrency/RatedCounter.cs
```diff
@@ -0,0 +1,160 @@
...
+            : base(rate, 86400 * 1000, name, logger)
+        {
+        }
+    }
+
+    // Leaky bucket counter
+    public abstract class RatedCounter
+    {
+        // At least 1 second duration
+        private const double MIN_TIME_UNIT = 1000;
```
timlaverty: _MS?

Services/Concurrency/RatedCounter.cs
```diff
@@ -0,0 +1,160 @@
...
+                var count = this.timestamps.Count;
+                if (count < this.eventsPerTimeUnit)
+                {
+                    this.lastTimestamp = now;
+                    this.timestamps.Enqueue(now);
+                    return false;
+                }
+
+                // Calculate when the next event is allowed to run, i.e. how long to pause
+                var nextTime = (long) (this.lastTimestamp + this.timeUnitLength + this.eventInterval * (count - this.eventsPerTimeUnit));
```
timlaverty: use now instead of this.lastTimestamp?

Services/Concurrency/RatedCounter.cs
```diff
@@ -0,0 +1,160 @@
...
+                pause = nextTime - now;
+
+                this.lastTimestamp = nextTime;
+                this.timestamps.Enqueue(nextTime);
+            }
+
+            // The caller is send too many events, if this happens you
+            // should consider redesigning the simulation logic to run
+            // slower, rather than relying purley on the counter
```
timlaverty: nit: purely

Services/DeviceClient.cs
```diff
@@ -41,27 +42,54 @@ public class DeviceClient : IDeviceClient
         private readonly Azure.Devices.Client.DeviceClient client;
         private readonly ILogger log;
         private readonly IoTHubProtocol protocol;
+        private readonly IRateLimiting rateLimiting;
         private readonly string deviceId;
         private readonly IScriptInterpreter scriptInterpreter;
 
         //used to create method pointers for the device for the IoTHub to callback to
+        // ReSharper disable once NotAccessedField.Local
```
timlaverty: nit: added comment from resharper?

Services/DeviceClient.cs
```diff
@@ -41,27 +42,54 @@ public class DeviceClient : IDeviceClient
...
             this.scriptInterpreter = scriptInterpreter;
+            this.connected = false;
         }
 
-        public IoTHubProtocol Protocol => this.protocol;
+        public async Task ConnectAsync()
+        {
+            if (this.client != null && !this.connected)
+            {
+                // TODO: HTTP clients don't "connect", find out how HTTP connections are measured and throttled
```
timlaverty: GitHub Issue?

Services/DeviceClient.cs
```diff
@@ -41,27 +42,54 @@ public class DeviceClient : IDeviceClient
...
+                await this.rateLimiting.LimitConnectionsAsync(() => this.client.OpenAsync());
```
timlaverty: I find passing the function pointer hard to read, we should discuss.

SimulationAgent/DependencyResolution.cs
```diff
@@ -72,14 +74,14 @@ private static void SetupCustomRules(ContainerBuilder builder)
             builder.RegisterType<Simulations>().As<ISimulations>().SingleInstance();
             builder.RegisterType<DeviceModels>().As<IDeviceModels>().SingleInstance();
             builder.RegisterType<Services.Devices>().As<IDevices>().SingleInstance();
+            builder.RegisterType<RateLimiting>().As<IRateLimiting>().SingleInstance();
```
timlaverty: previously you have a todo for ensuring this is a singleton - is it created anywhere explicitly?  Using DI you're protecting it already it looks like.

SimulationAgent/Program.cs
```diff
@@ -26,7 +26,13 @@ private static void PrintBootstrapInfo(IContainer container)
             var config = container.Resolve<IConfig>();
             logger.Info("Simulation agent started", () => new { Uptime.ProcessId });
             logger.Info("Device Models folder: " + config.ServicesConfig.DeviceModelsFolder, () => { });
-            logger.Info("Scripts folder:      " + config.ServicesConfig.DeviceModelsScriptsFolder, () => { });
+            logger.Info("Scripts folder:       " + config.ServicesConfig.DeviceModelsScriptsFolder, () => { });
+
+            logger.Info("Connections per sec:  " + config.ServicesConfig.RateLimiting.ConnectionsPerSecond, () => { });
```
timlaverty: should we wrap these under config entries for S1/S2/S3 with 1 Unit?

SimulationAgent/Runtime/Config.cs
```diff
@@ -46,13 +53,23 @@ public Config(IConfigData configData)
                                     "value in the 'appsettings.ini' configuration file.");
             }
 
+            var limitsConf = new RateLimitingConfiguration
+            {
+                ConnectionsPerSecond = configData.GetDouble(CONNECTIONS_FREQUENCY_LIMIT_KEY),
+                RegistryOperationsPerMinute = configData.GetDouble(REGISTRYOPS_FREQUENCY_LIMIT_KEY),
+                MessagesPerDay = configData.GetDouble(MESSAGES_FREQUENCY_LIMIT_KEY),
```
timlaverty: nit: whole number/integer for daily message cap

SimulationAgent/Runtime/Config.cs
```diff
@@ -46,13 +53,23 @@ public Config(IConfigData configData)
                                     "value in the 'appsettings.ini' configuration file.");
             }
 
+            var limitsConf = new RateLimitingConfiguration
+            {
+                ConnectionsPerSecond = configData.GetDouble(CONNECTIONS_FREQUENCY_LIMIT_KEY),
```
timlaverty: super nit: integer for connections

SimulationAgent/Runtime/Config.cs
```diff
@@ -46,13 +53,23 @@ public Config(IConfigData configData)
...
+                RegistryOperationsPerMinute = configData.GetDouble(REGISTRYOPS_FREQUENCY_LIMIT_KEY),
+                MessagesPerDay = configData.GetDouble(MESSAGES_FREQUENCY_LIMIT_KEY),
+                TwinReadsPerSecond = configData.GetDouble(TWIN_READS_FREQUENCY_LIMIT_KEY),
+                TwinWritesPerSecond = configData.GetDouble(TWIN_WRITES_FREQUENCY_LIMIT_KEY)
```
timlaverty: Where is the throttle config for Device-to-cloud sends?  For S1 == Higher of 100/sec or 12/sec/unit 

https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling

Services/Devices.cs
```diff
@@ -76,19 +91,37 @@ public IDeviceClient GetClient(Device device, IoTHubProtocol protocol, IScriptIn
...
-                await Task.WhenAll(device, twin);
+                Azure.Devices.Device device = null;
+                Twin twin = null;
+
+                if (loadTwin)
+                {
+                    var deviceTask = this.rateLimiting.LimitRegistryOperationsAsync(
+                        () => this.registry.GetDeviceAsync(deviceId));
+
+                    var twinTask = this.rateLimiting.LimitTwinReadsAsync(
```
timlaverty: very nice having this split out.

SimulationAgent/Simulation/DeviceStatusLogic/Connect.cs
```diff
@@ -73,13 +73,14 @@ public void Run(object context)
...
+                        var task = this.devices.GetOrCreateAsync(this.deviceId, false);
                         task.Wait((int) connectionTimeout.TotalMilliseconds, actor.CancellationToken);
                         var device = task.Result;
 
                         this.log.Debug("Device credentials retrieved", () => new { this.deviceId });
 
                         actor.Client = this.devices.GetClient(device, this.protocol.Value, this.scriptInterpreter);
+                        actor.Client.ConnectAsync().Wait(connectionTimeout);
 
                         // Device Twin properties can be set only over MQTT, so we need a dedicated client
```
timlaverty: Twin reads are supported over AMQP with the latest SDK - we should prep to upgrade.  Can you add a todo for it?

SimulationAgent/appsettings.ini
```diff
@@ -5,12 +5,37 @@ iothub_connstring = "${PCS_IOTHUB_CONNSTRING}"
...
+# F1:  10/sec
+twin_reads_per_second = 10
+# S3:  50/sec/unit
+# S2:   1/sec/unit  (10/sec minimum)
+# S1:  10/sec
+# F1:  10/sec
+twin_writes_per_second = 10
+# S3:  6000/sec/unit
+# S2:   120/sec/unit
+# S1:    12/sec/unit (100/sec minimum)
```
timlaverty: nit, higher of 12/sec/unit or 100/sec (not 100/sec min)

https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling

SimulationAgent/appsettings.ini
```diff
@@ -5,12 +5,37 @@ iothub_connstring = "${PCS_IOTHUB_CONNSTRING}"
...
+# S1:   100/min/unit (= 1.67/sec/unit)
+# F1:   100/min      (= 1.67/sec)
+registry_operations_per_minute = 100
+# S3:  50/sec/unit
+# S2:   1/sec/unit  (10/sec minimum)
+# S1:  10/sec
+# F1:  10/sec
+twin_reads_per_second = 10
+# S3:  50/sec/unit
+# S2:   1/sec/unit  (10/sec minimum)
```
timlaverty: nit, not a minimum?

SimulationAgent/appsettings.ini
```diff
@@ -5,12 +5,37 @@ iothub_connstring = "${PCS_IOTHUB_CONNSTRING}"
 device_models_folder = ./data/devicemodels/
 device_models_scripts_folder = ./data/devicemodels/scripts/
 
+[DeviceSimulationService:RateLimits]
+# S3:  5000/min/unit (= 83.3/sec/unit)
+# S2:   100/min/unit (= 1.67/sec/unit)
+# S1:   100/min/unit (= 1.67/sec/unit)
+# F1:   100/min      (= 1.67/sec)
+registry_operations_per_minute = 100
```
timlaverty: I'd like to limit to 1/2 or 3/4 the available limits to allow for clients outside the simulator leeway instead of running the simulator @ capacity.
e.g. registry_operations_per_minute = 50 or 75

SimulationAgent/appsettings.ini
```diff
@@ -5,12 +5,37 @@ iothub_connstring = "${PCS_IOTHUB_CONNSTRING}"
...
+# S3:  50/sec/unit
+# S2:   1/sec/unit  (10/sec minimum)
+# S1:  10/sec
+# F1:  10/sec
+twin_reads_per_second = 10
```
timlaverty: 5

SimulationAgent/appsettings.ini
```diff
@@ -5,12 +5,37 @@ iothub_connstring = "${PCS_IOTHUB_CONNSTRING}"
...
+# S3:  50/sec/unit
+# S2:   1/sec/unit  (10/sec minimum)
+# S1:  10/sec
+# F1:  10/sec
+twin_writes_per_second = 10
```
timlaverty: 5

SimulationAgent/appsettings.ini
```diff
@@ -5,12 +5,37 @@ iothub_connstring = "${PCS_IOTHUB_CONNSTRING}"
...
+# S3:  6000/sec/unit
+# S2:   120/sec/unit
+# S1:    12/sec/unit (100/sec minimum)
+# F1:   100/sec
+device_connections_per_second = 100
```
timlaverty: 50

SimulationAgent/appsettings.ini
```diff
@@ -5,12 +5,37 @@ iothub_connstring = "${PCS_IOTHUB_CONNSTRING}"
...
+# S3:  300M/day/unit (= 208333.33/min/day)
+# S2:    6M/day/unit (=   4166.67/min/unit)
+# S1:  400k/day/unit (=    277.78/min/unit)
+# F1:  8000/day      (=      5.56/min)
+messages_per_day = 400000
```
timlaverty: 300k?  we should leave room for physical devices.

Services.Test/Concurrency/PerSecondCounterTest.cs
dluc: we had a preexisting "concurrency" namespace which seemed good enough. I'm ok moving the code into a dedicated space if you say, however I think it's still in the same space of problems

Services.Test/Concurrency/PerSecondCounterTest.cs
dluc: good point, I'll make the change

Services/Concurrency/RateLimiting.cs
dluc: will add

Services/Concurrency/RatedCounter.cs
dluc: I'd have to add "ms" or "msecs" to all the vars, which I did at some point and made the code harder to read. I prefer the assumption that all the timing values are expressed in msecs by definition, see also the comment on line 46

Services/Concurrency/RatedCounter.cs
dluc: `this.lastTimestamp` is the time when the last event in the queue can be executed, which is most likely in the future

Services/DeviceClient.cs
dluc: yes, resharper keeps warning about this var not being used (something we discussed in another PR), which was added on purpose (you did the work to support methods a while back). The comment is there so that I don't get resharper warnings

Services/DeviceClient.cs
dluc: sure - it's a common pattern in AOP, usually hidden by means of runtime or compilation tools (e.g. https://www.postsharp.net/aop.net).
One alternative would be using 2 lines of code, one to increase the counter and pause, the next to run the command, however that is more error prone in my opinion, because one must be sure both lines are present and in the right order.

SimulationAgent/DependencyResolution.cs
dluc: yes and no, this is kind of tricky. When everything works well, DI will ensure this is a singleton. However, when the singleton container is disposed, this singleton is disposed too which is a bummer because we lose all the counters. Now, that happens only when there are problems while bootstrapping the app, e.g. storage is down, so it's not a big deal right now, but that TODO is to remind me of this issue. I think having a reference in Program.cs will make sure the counters are never reset (something to investigate later).

SimulationAgent/Program.cs
dluc: I thought about having configurations per SKU, but I think that's now how people will consume the service. I added the comments in application.ini to explain how to calculate values, but I prefer the fully customizable values, that don't have to match exactly the SKUs. For example, I might want to reduce the number of twin ops more than other limits, because of Web UI (or other UIs)

SimulationAgent/Runtime/Config.cs
dluc: I rather not, many values are expressed as real numbers also in our docs, I consider this as a feature (e.g. 2.5 connections per second is a possible scenario)

SimulationAgent/Runtime/Config.cs
dluc: I'd prefer keeping all the values as real numbers for consistency and easier documentation

SimulationAgent/Runtime/Config.cs
dluc: we don't use c2d messages in the simulation, right?

SimulationAgent/Simulation/DeviceStatusLogic/Connect.cs
dluc: 👍  we need to test the perf and network reliability, will add task

SimulationAgent/appsettings.ini
dluc: sure, let's recap all the limits in a separate task and leave these here for this PR?

SimulationAgent/appsettings.ini
dluc: ? - S2 allows at least 10/sec, should I rephrase it?

Services/Concurrency/RateLimiting.cs
dluc: done (will push code)

Services/DeviceClient.cs
dluc: added comment pointing to https://github.com/Azure/device-simulation-dotnet/issues/85

SimulationAgent/DependencyResolution.cs
dluc: added https://github.com/Azure/device-simulation-dotnet/issues/79

SimulationAgent/appsettings.ini
dluc: https://github.com/Azure/device-simulation-dotnet/issues/93

By the way the values in the repository might differ from what we want to use from deployment to deployment, and for other solutions

SimulationAgent/appsettings.ini
dluc: changed to "higher of 10/sec and 1/sec/unit"

SimulationAgent/appsettings.ini
dluc: changed to "higher of 100/sec and 12/sec/unit"

SimulationAgent/appsettings.ini
dluc: see https://github.com/Azure/device-simulation-dotnet/issues/93

Services.Test/Concurrency/PerSecondCounterTest.cs
timlaverty: I'm ok w/ it.  I think of it as a Throttling specific set of problems but I can buy into that being a concurrency problem (multiple threads access the hub concurrently causing problems if not adjudicated).

Services/DeviceClient.cs
timlaverty: Got it.  I will clean this up in my next PR... it's less clean to see the warnings and to have this in the repo.

Services/Devices.cs
```diff
@@ -42,7 +52,8 @@ public class Devices : IDevices
         {
```
timlaverty: Who uses this? A test? Do we need it?

Services/Devices.cs
```diff
@@ -42,7 +52,8 @@ public class Devices : IDevices
...
             try
             {
-                await this.registry.GetDeviceAsync("healthcheck");
+                await this.rateLimiting.LimitRegistryOperationsAsync(
+                    () => this.registry.GetDeviceAsync("healthcheck"));
```
timlaverty: I'm assuming the device will always return null and will not throw, yes?

SimulationAgent/DependencyResolution.cs
timlaverty: >> I think having a reference in Program.cs will make sure the counters are never reset (something to investigate later).

This makes a lot of sense to me - not critical for now though.

SimulationAgent/Program.cs
timlaverty: Got it - this makes sense WRT customizing for the # of clients.  If you have a low # of people on the cloud side accessing the data I might configure simulation to run closer to the hub limits.

SimulationAgent/Properties/launchSettings.json
```diff
@@ -8,4 +8,4 @@
       }
     }
   }
-}
\ No newline at end of file
+}
```
timlaverty: weird, not sure how this worked w/o the match

SimulationAgent/Runtime/Config.cs
timlaverty: From a hub side only integer works.  I can see a world where you are going so far as to fine tune as you're suggesting.  I'll be surprised though :).

Services/DeviceClient.cs
dluc: fixed in the other PR 👍

Services/Devices.cs
dluc: removed in the other PR

Services/Devices.cs
dluc: interesting, this doesn't look right, I'll create a task to investigate

dluc: https://github.com/Azure/device-simulation-dotnet/issues/97

SimulationAgent/Runtime/Config.cs
dluc: changed to Int in the other PR

dluc:
Other PR merged into this, so we should be ready to merge to master and release the new build

PR #99, opened 2017-10-20
Set method result to empty string @ execution completion.  

timlaverty:
…ion handler to callback function called by hub.

# Type of change? 

- [ x] Bug fix
- [ ] New feature
- [ ] Enhancement

Set method result for firmwareupdate and reboot to empty string @ execution completion.  Add exception handler to function pointer that the hub calls for method execution.
...

**Checklist:**

- [x ] All tests passed
- [ x] The code follows the code style and conventions of this project

### DIFF
--- a/Services/DeviceMethods.cs
+++ b/Services/DeviceMethods.cs
@@ -79,21 +79,27 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
 
         public Task<MethodResponse> ExecuteMethodAsync(MethodRequest methodRequest, object userContext)
         {
-            this.log.Info("Creating task to execute method with json payload.", () => new
+            try
             {
-                this.deviceId,
-                methodName = methodRequest.Name,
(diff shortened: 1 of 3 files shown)


Services/DeviceMethods.cs
```diff
@@ -79,21 +79,27 @@ public class DeviceMethods : IDeviceMethods
...
+                });
+
+                // Kick the method off on a separate thread & immediately return
+                // Not immediately returning would block the client connection to the hub
+                var t = Task.Run(() => this.MethodExecution(methodRequest));
 
-            // Kick the method off on a separate thread & immediately return
-            // Not immediately returning would block the client connection to the hub
-            var t = Task.Run(() => this.MethodExecution(methodRequest));
+                return Task.FromResult(new MethodResponse((int) HttpStatusCode.OK));
```
dluc: so we won't return a message like earlier "Executed Method: ...". Was that an issue?

PR #103, opened 2017-10-27
Update device models, allow to demo chillers fix/break

dluc:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

* Update the chiller device models to allow fixing/breaking them during a demo
* Allow methods to suspend the telemetry, e.g. during a reboot
* Refactor the scripting engine, add some more functions to manage the methods execution progress
* Update the device models with the specs provided (add also the Tags, even if they are not supported yet)

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/103)

### DIFF
--- a/Services/Simulation/JavascriptInterpreter.cs
+++ b/Services/Simulation/JavascriptInterpreter.cs
@@ -53,19 +53,28 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Simulation
             // logging into the service logs
             engine.SetValue("log", new Action<object>(this.JsLog));
 
-            //register callback for state updates
-            engine.SetValue("updateState", new Action<JsValue>(this.UpdateState));
+            // Register callback for state updates
+            engine.SetValue("updateState", new Action<JsValue>(this.JsMergeState));
+            engine.SetValue("reportMethodProgress", new Action<string>(this.JsReportMethodProgress));
 
-            //register sleep function for javascript use
-            engine.SetValue("sleep", new Action<int>(this.Sleep));
+            // Register sleep function for javascript use
+            engine.SetValue("sleep", new Action<int>(this.JsSleep));
+
+            // Register functions to enable/disable internal state simulation
+            engine.SetValue("enableSensorSimulation", new Action(this.JsEnableSensorSimulation));
+            engine.SetValue("disableSensorSimulation", new Action(this.JsDisableSensorSimulation));
+
+            // Register functions to enable/disable telemetry
+            engine.SetValue("enableTelemetry", new Action(this.JsEnableTelemetry));
+            engine.SetValue("disableTelemetry", new Action(this.JsDisableTelemetry));
 
             var sourceCode = this.LoadScript(filename);
             this.log.Debug("Executing JS function", () => new { filename });
 
             try
             {
-                var output = engine.Execute(sourceCode).Invoke("main", context, this.deviceState);
-                var result = this.JsValueToDictionary(output);
+                JsValue output = engine.Execute(sourceCode).Invoke("main", context, this.deviceState);
+                Dictionary<string, object> result = this.JsValueToDictionary(output);
                 this.log.Debug("JS function success", () => new { filename, result });
                 return result;
             }
@@ -145,14 +154,14 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Simulation
             this.log.Debug("Log from JS", () => new { data });
         }
 
-        private void Sleep(int timeInMs)
+        private void JsSleep(int timeInMs)
         {
             Task.Delay(timeInMs).Wait();
         }
 
         // TODO: Move this out of the scriptinterpreter class into DeviceClient to keep this class stateless
         //       https://github.com/Azure/device-simulation-dotnet/issues/45
-        private void UpdateState(JsValue data)
+        private void JsMergeState(JsValue data)
         {
             string key;
             object value;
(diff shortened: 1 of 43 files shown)


Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -53,19 +53,28 @@ public class JavascriptInterpreter : IJavascriptInterpreter
             // logging into the service logs
             engine.SetValue("log", new Action<object>(this.JsLog));
 
-            //register callback for state updates
-            engine.SetValue("updateState", new Action<JsValue>(this.UpdateState));
+            // Register callback for state updates
+            engine.SetValue("updateState", new Action<JsValue>(this.JsMergeState));
+            engine.SetValue("reportMethodProgress", new Action<string>(this.JsReportMethodProgress));
```
timlaverty: I'm not sure it's worth changing method progress.  We don't use the method progression reported properties in the twin from the UI right now.  Method progress itself needs redesigned and the sdk support for individual device execution of methods needs evaluated.

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -53,19 +53,28 @@ public class JavascriptInterpreter : IJavascriptInterpreter
...
 
-            //register sleep function for javascript use
-            engine.SetValue("sleep", new Action<int>(this.Sleep));
+            // Register sleep function for javascript use
+            engine.SetValue("sleep", new Action<int>(this.JsSleep));
+
+            // Register functions to enable/disable internal state simulation
+            engine.SetValue("enableSensorSimulation", new Action(this.JsEnableSensorSimulation));
```
timlaverty: I pictured us turning sensor simulation off at an individual sensor level, not for all sensors.

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -53,19 +53,28 @@ public class JavascriptInterpreter : IJavascriptInterpreter
...
+            engine.SetValue("disableSensorSimulation", new Action(this.JsDisableSensorSimulation));
+
+            // Register functions to enable/disable telemetry
```
timlaverty: #supernit: "...functions to enable/disable sending telemetry to the hub"

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -145,14 +154,14 @@ private void JsLog(object data)
             this.log.Debug("Log from JS", () => new { data });
         }
 
-        private void Sleep(int timeInMs)
+        private void JsSleep(int timeInMs)
```
timlaverty: Could be used outside of Js?

Services/Simulation/ScriptInterpreter.cs
```diff
@@ -46,5 +50,40 @@ public class ScriptInterpreter : IScriptInterpreter
...
         }
+
+        public static void AddMethodStatusProperty(Dictionary<string, object> state)
+        {
+            state[DEVICE_METHOD_STATUS] = "";
+        }
+
+        public static void EnableSensorSimulation(Dictionary<string, object> state)
+        {
+            state.Add(SENSOR_SIMULATION_ENABLED, true);
```
timlaverty: should you check to see if it exists & set it if yes?  ...adding it if it doesn't

Services/data/devicemodels/scripts/EmergencyValveRelease-method.js
```diff
@@ -2,32 +2,38 @@
...
+
+    disableSensorSimulation();
+
     state.pressure = 150;
-    state.CalculateRandomizedTelemetry = true;
-    // update the state to 150
     updateState(state);
+    log("Pressure decreased to " + state.pressure);
+
+    enableSensorSimulation();
```
timlaverty: I don't think you want to enable it again as the simulation calculation will overwrite the pressure value you just set.  I think you want to keep it disabled.

Services/data/devicemodels/scripts/IncreasePressure-method.js
```diff
@@ -22,11 +21,40 @@ var state = {
...
+    sleep(1000);
+
     state.pressure = 250;
-    state.CalculateRandomizedTelemetry = false;
-    // update the state to 250
     updateState(state);
+    log("Pressure increased to " + state.pressure);
+    sleep(1000);
+
+    enableSensorSimulation();
```
timlaverty: probably want to leave it disabled so more alarms aren't generated.

SimulationAgent/Simulation/DeviceStatusLogic/UpdateDeviceState.cs
```diff
@@ -108,34 +106,35 @@ private void RunInternal()
...
-                // until the correlating function has been called; e.g. when increasepressure is called, don't write
-                // telemetry until decreasepressure is called for that property.
-                this.log.Debug("Checking for the need to compute new telemetry",
+                // until the correlating function has been called; e.g. when increasepressure is called,
+                // don't write telemetry until decreasepressure is called for that property
+                this.log.Debug("Checking for the need to compute new sensors state",
                     () => new { this.deviceId, deviceState = actor.DeviceState });
-                if ((bool) actor.DeviceState[CALC_TELEMETRY])
+
+                if (ScriptInterpreter.IsSensorSimulationEnabled(actor.DeviceState))
```
timlaverty: it seems odd that the scriptinterpreter would own whether sensor simulation is enabled.  I'd rather keep this in the actor.

dluc: will refactor, clean up will make the PR a little bigger

dluc:
the PR is getting too big and deserves some more cooking time, I will open a new one removing all the cleanup and refactoring

PR #125, opened 2017-11-15
Disable twin read write by default and add config option to enable

jillcary:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

This change makes it so that, by default, twin read/writes and methods are disabled to reduce the number of threads and open connections. 

- configuration option in appsettings.ini file
- false by default
- disables calls to read and write to twin

**Checklist:**

- [x ] All tests passed

### DIFF
--- a/Services/Devices.cs
+++ b/Services/Devices.cs
@@ -190,13 +190,11 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
                     () => new { device.Id, DeviceTwin.SIMULATED_TAG_KEY, DeviceTwin.SIMULATED_TAG_VALUE });
                 twin.Tags[DeviceTwin.SIMULATED_TAG_KEY] = DeviceTwin.SIMULATED_TAG_VALUE;
 
-                if (this.twinReadsWritesEnabled)
-                {
-                    // TODO: when not discarding the twin, use the right ETag and manage conflicts
-                    //       https://github.com/Azure/device-simulation-dotnet/issues/83
-                    twin = await this.rateLimiting.LimitTwinWritesAsync(
-                        () => this.registry.UpdateTwinAsync(device.Id, twin, "*", cancellationToken));
-                }
+
+                // TODO: when not discarding the twin, use the right ETag and manage conflicts
+                //       https://github.com/Azure/device-simulation-dotnet/issues/83
+                twin = await this.rateLimiting.LimitTwinWritesAsync(
+                    () => this.registry.UpdateTwinAsync(device.Id, twin, "*", cancellationToken));
 
                 return new Device(device, twin, this.ioTHubHostName);
             }
--- a/Services/Devices.cs
+++ b/Services/Devices.cs
@@ -85,10 +85,11 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
 
         public async Task<Device> GetOrCreateAsync(string deviceId, bool loadTwin, CancellationToken cancellationToken)
         {
+
+            if (!this.twinReadsWritesEnabled) { loadTwin = false; }
+
             try
             {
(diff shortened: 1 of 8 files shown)


Services/Devices.cs
```diff
@@ -85,6 +87,8 @@ public class Devices : IDevices
         {
             try
             {
+                if (!this.twinReadsWritesEnabled) { loadTwin = false; }
```
dluc: nit: before the `try`, this is not supposed to throw

SimulationAgent/Runtime/Config.cs
```diff
@@ -20,6 +20,8 @@ public class Config : IConfig
         private const string DEVICE_MODELS_SCRIPTS_FOLDER_KEY = APPLICATION_KEY + "device_models_scripts_folder";
         private const string IOTHUB_CONNSTRING_KEY = APPLICATION_KEY + "iothub_connstring";
 
+        private const string TWIN_READS_WRITES_ENABLED_KEY = APPLICATION_KEY + "twin_read_write_enabled";
```
dluc: could you use singular or plural for both the C# constant and the config prop?

dluc: (+1 for singular, "read-write" is quite common)

SimulationAgent/Simulation/DeviceActor.cs
```diff
@@ -207,7 +213,11 @@ public IDeviceActor Setup(DeviceModel deviceModel, int position)
             this.updateDeviceStateLogic.Setup(this.deviceId, deviceModel, this);
             this.deviceBootstrapLogic.Setup(this.deviceId, deviceModel, this);
             this.sendTelemetryLogic.Setup(this.deviceId, deviceModel, this);
-            this.updateReportedPropertiesLogic.Setup(this.deviceId, deviceModel, this);
+
+            if (this.twinReadsWritesEnabled)
```
dluc: I would move this logic inside of `UpdateReportedProperties`. DI will instantiate that class anyway, so we need to make sure that the object won't do anything

SimulationAgent/appsettings.ini
```diff
@@ -4,6 +4,10 @@ iothub_connstring = "${PCS_IOTHUB_CONNSTRING}"
 # which can be mounted to inject custom device models and scripts.
 device_models_folder = ./data/devicemodels/
 device_models_scripts_folder = ./data/devicemodels/scripts/
+; Disable twin read/writes and methods, to reduce the number of threads
+; and open connections.
+twin_read_write_enabled = false;
```
dluc: I would use an optional Env Var here:
1. if the env var is empty, default to True (RMv2 behavior)
2. if the env var is available, it can be any of T,F,true,false,on,off,0,1,-1 (same as the implementation of "auth_required" in the webservice)

SimulationAgent/Simulation/DeviceActor.cs
```diff
@@ -305,9 +315,13 @@ public void MoveNext()
 
                 case Status.UpdatingReportedProperties:
                     this.ActorStatus = nextStatus;
-                    this.updateReportedPropertiesLogic.Start();
-                    // Note: at this point both UpdatingDeviceState
-                    //       and UpdatingReportedProperties should be running
+                    // only update twin if enabled in config
+                    if (this.twinReadsWritesEnabled)
```
dluc: ditto, not needed

SimulationAgent/appsettings.ini
jillcary: If we default to true, then the service will have twin read/write enabled by default. Don't we want it to default to false for performance reasons or is this something the CLI will set on deployment?

jillcary: Update: This needs to be backwards compatible with RMv2, so we should default to true and the CLI will set the config value of false.

PR #138, opened 2017-11-29
New web service APIs

dluc:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [x] Breaking change (breaks backward compatibility)

* New API to support multiple scripts 
* New API to support custom sensors
* New API to support custom telemetry frequency
* New API to expose whether a hub is configured, i.e. provisioned and the connection string is in the config file
* Fix exception occurring when there is no connection string

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/138)


Services/Devices.cs
```diff
@@ -71,14 +72,16 @@ public class Devices : IDevices
...
+            this.setupDone = false;
         }
-        
+
         /// <summary>
         /// Ping the registry to see if the connection is healthy
         /// </summary>
         public async Task<Tuple<bool, string>> PingRegistryAsync()
         {
+            this.SetupHub();
```
timlaverty: Why move this outside the constructor?  It looks like every function needs it so that seems like the right spot.

Services/Devices.cs
```diff
@@ -198,6 +209,15 @@ public void SetCurrentIotHub()
             this.log.Info("Selected active IoT Hub for devices", () => new { this.ioTHubHostName });
         }
 
+        private void SetupHub()
+        {
+            if (this.setupDone) return;
+
+            this.SetCurrentIotHub();
```
timlaverty: Is this OK to be reentrant; i.e. is anything shared across threads?  It has a race condition for if methods are called in parallel.

Services/data/devicemodels/chiller-01.json
```diff
@@ -16,11 +16,13 @@
             "pressure_unit": "psig",
             "simulation_state": "normal_pressure"
         },
-        "Script": {
-            "Type": "javascript",
-            "Path": "chiller-01-state.js",
-            "Interval": "00:00:05"
-        }
+        "Interval": "00:00:10",
```
timlaverty: why change the structure?  This would allow multiple state scripts that run @ different intervals but I'm not sure we will need this.

WebService/v1/Controllers/StatusController.cs
```diff
@@ -41,19 +45,42 @@ public sealed class StatusController : Controller
...
+                    iotHubStatus = await this.devices.PingRegistryAsync();
+                    if (!iotHubStatus.Item1)
+                    {
+                        statusIsOk = false;
+                        errors.Add("Unable to use Azure IoT Hub service");
+                    }
+                }
+                else
+                {
+                    iotHubStatus = new Tuple<bool, string>(false, "not configured");
```
timlaverty: nit: "IoTHub connection string not configured"

WebService/v1/Models/SimulationApiModel/DeviceModelApiModelOverride.cs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApiModel
+{
+    // SEE: <DeviceModelApiModel> for the original fields being overridden
```
timlaverty: this class seems odd to me - why wouldn't you subclass the devicemodelapimodel class if that's what you're actually doing?

WebService/v1/Models/SimulationApiModel/DeviceModelSimulationOverride.cs
```diff
@@ -0,0 +1,70 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApiModel
+{
+    // SEE: <DeviceModelApiModel.DeviceModelSimulation> for the original fields being overridden
```
timlaverty: same comment as devicemodelapimodel subclass above.

WebService/v1/Controllers/StatusController.cs
```diff
@@ -17,20 +18,23 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
...
+            IServicesConfig servicesConfig)
         {
             this.devices = devices;
             this.storage = storage;
             this.simulations = simulations;
             this.log = logger;
+            this.servicesConfig = servicesConfig;
         }
 
         [HttpGet]
```
jillcary: Split method into sub-methods for readability/debugging.

WebService/v1/Models/DeviceModelApiModel/DeviceModelApiModel.cs
```diff
@@ -0,0 +1,83 @@
...
+        public IDictionary<string, DeviceModelSimulationScript> CloudToDeviceMethods { get; set; }
+
+        [JsonProperty(PropertyName = "$metadata", Order = 1000)]
+        public IDictionary<string, string> Metadata => new Dictionary<string, string>
+        {
+            { "$type", "DeviceModel;" + v1.Version.NUMBER },
+            { "$uri", "/" + v1.Version.PATH + "/devicemodels/" + this.Id }
+        };
+
+        public DeviceModelApiModel()
```
jillcary: Are all values accounted for here? Initialize to null or default if possible.

WebService/v1/Models/DeviceModelApiModel/DeviceModelApiModel.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            this.Simulation = new DeviceModelSimulation();
+            this.Telemetry = new List<DeviceModelTelemetry>();
+            this.Properties = new Dictionary<string, object>();
+            this.CloudToDeviceMethods = new Dictionary<string, DeviceModelSimulationScript>();
+        }
+
+        /// <summary>Map a service model to the corresponding API model</summary>
```
jillcary: TODO: use builder pattern similar to simulation

WebService/v1/Models/DeviceModelApiModel/DeviceModelSimulation.cs
```diff
@@ -0,0 +1,38 @@
...
+        public List<DeviceModelSimulationScript> SimulationScripts { get; set; }
+
+        public DeviceModelSimulation()
+        {
+            this.Initial = new Dictionary<string, object>();
+            this.Interval = null;
+            this.SimulationScripts = new List<DeviceModelSimulationScript>();
+        }
+
+        /// <summary>Map a service model to the corresponding API model</summary>
```
jillcary: TODO: use builder pattern

WebService/v1/Models/DeviceModelApiModel/DeviceModelSimulationScript.cs
```diff
@@ -0,0 +1,31 @@
...
+        [JsonProperty(PropertyName = "Path")]
+        public string Path { get; set; }
+
+        public DeviceModelSimulationScript()
+        {
+            this.Type = "javascript";
+            this.Path = "scripts" + System.IO.Path.DirectorySeparatorChar;
+        }
+
+        /// <summary>Map a service model to the corresponding API model</summary>
```
jillcary: TODO: builder pattern

WebService/v1/Models/DeviceModelApiModel/DeviceModelTelemetry.cs
```diff
@@ -0,0 +1,36 @@
...
+        public DeviceModelTelemetryMessageSchema MessageSchema { get; set; }
+
+        public DeviceModelTelemetry()
+        {
+            this.Interval = "00:00:00";
+            this.MessageTemplate = string.Empty;
+            this.MessageSchema = new DeviceModelTelemetryMessageSchema();
+        }
+
+        /// <summary>Map a service model to the corresponding API model</summary>
```
jillcary: TODO: builder pattern

WebService/v1/Models/DeviceModelApiModel/DeviceModelTelemetryMessageSchema.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public DeviceModelTelemetryMessageSchema()
+        {
+            this.Name = string.Empty;
+            this.Format = "JSON";
+            this.Fields = new Dictionary<string, string>();
+        }
+
+        /// <summary>Map a service model to the corresponding API model</summary>
+        public DeviceModelTelemetryMessageSchema(DeviceModel.DeviceModelMessageSchema schema) : this()
```
jillcary: TODO: builder pattern

WebService/v1/Models/DeviceModelListApiModel.cs
```diff
@@ -21,14 +21,14 @@ public class DeviceModelListApiModel
 
         public DeviceModelListApiModel()
         {
-            this.Items = new List<DeviceModelApiModel>();
+            this.Items = new List<DeviceModelApiModel.DeviceModelApiModel>();
         }
 
         /// <summary>Map a service model to the corresponding API model</summary>
         public DeviceModelListApiModel(IEnumerable<Services.Models.DeviceModel> deviceModels)
```
jillcary: TODO: builder pattern

WebService/v1/Models/SimulationApiModel/DeviceModelSimulationScriptOverride.cs
```diff
@@ -0,0 +1,72 @@
...
+        [JsonProperty(PropertyName = "Type", NullValueHandling = NullValueHandling.Ignore)]
+        public string Type { get; set; }
+
+        // Optional, used to change the script used
+        [JsonProperty(PropertyName = "Path", NullValueHandling = NullValueHandling.Ignore)]
+        public string Path { get; set; }
+
+        // Optional, used to provide input parameters to the script
+        [JsonProperty(PropertyName = "Params", NullValueHandling = NullValueHandling.Ignore)]
+        public object Params { get; set; }
```
jillcary: Consider adding validation using a method provided by the "path" function.

WebService/v1/Controllers/StatusController.cs
dluc: 👍

WebService/v1/Controllers/StatusController.cs
dluc: :+1:

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -276,17 +280,36 @@ Content-Type: application/json; charset=utf-8
...
+a device model. Note that the order of the overrides is important.
+
+The example creates a simulation using 4 device models: *truck-01*,
+*truck-02*, *elevator-01*, *elevator-02*.
+Only the frequency of *truck-01* is customized, while the others models use
+the default settings, defined in their respective JSON files (*truck-02.json*,
+*elevator-01.json*, *elevator-02.json*).
+
+The *truck-01* device model defines 3 distinct telemetry messages, and
+the example shows how to set a custom frequency for each one, in order
```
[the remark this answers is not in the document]
dluc: :+1:

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -321,13 +344,30 @@ Content-Type: application/json; charset=utf-8
...
+### Example: Customize scripts (custom sensors)
+
+The following example shows how to customize the scripts used to generate
+the virtual sensors state, and the content of the telemetry messages.
+
+Similarly to the previous example, the simulation uses multiple models,
+but only one (the third in this case) is customized.
+
+Some important notes:
+* when customizing the **telemetry content**, all the details must be included:
```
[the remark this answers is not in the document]
dluc: :+1:

WebService/v1/Models/DeviceModelApiModel/DeviceModelApiModel.cs
dluc: :+1:

WebService/v1/Models/SimulationApiModel/DeviceModelSimulationScriptOverride.cs
dluc: :+1: added TODO and github issue

WebService/v1/Models/DeviceModelApiModel/DeviceModelApiModel.cs
dluc: :+1:

WebService/v1/Models/DeviceModelApiModel/DeviceModelSimulation.cs
dluc: 👍

WebService/v1/Models/DeviceModelApiModel/DeviceModelSimulationScript.cs
dluc: 👍

WebService/v1/Models/DeviceModelApiModel/DeviceModelTelemetry.cs
dluc: 👍

WebService/v1/Models/DeviceModelApiModel/DeviceModelTelemetryMessageSchema.cs
dluc: 👍

WebService/v1/Models/DeviceModelListApiModel.cs
dluc: 👍

WebService/v1/Models/SimulationApiModel/DeviceModelApiModelOverride.cs
dluc: the original class has a lot of properties that cannot be overridden and would show up in here even if they cannot be overridden in the business logic. Also all the initialization is different, this class properties are all null by default. There's also a coupling problem, where changes to one model would affect the other, with a higher risk of breaking the API in case of changes. A separate class gives in my opinion better control on the features provided (i.e. what can be overridden).

WebService/v1/Models/SimulationApiModel/DeviceModelSimulationOverride.cs
dluc: replied in the other comment

Services/Devices.cs
dluc: the function can throw an exception, which is fine when the exception occurs in a method. If the exception occurs in the ctor instead, it breaks DI, for instance the StatusController would be unable to report that IoTHub is not working. I added a comment in the code to clarify this. Generally speaking, we should avoid dealing with external dependencies in the constructors, for the same reason, we can use lazy initialization or this kind of construct.

Services/data/devicemodels/chiller-01.json
dluc: While the feature would be interesting, we don't have a use case now, and the scheduling implementation is not trivial, so I decided to leave this out for the time being. Note that we're not dropping the feature, because before only 1 script was allowed.

PR #141, opened 2017-11-30
Use custom frequency for telemetry and device state

dluc:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Apply the user customizations to the device models, e.g. custom telemetry frequency and custom device state simulation frequency.
Fix a regression affecting device models with multiple telemetry messages.

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/141)

### DIFF
--- a/Services/DeviceModels.cs
+++ b/Services/DeviceModels.cs
@@ -18,6 +18,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
     {
         IEnumerable<DeviceModel> GetList();
         DeviceModel Get(string id);
+        DeviceModel OverrideDeviceModel(DeviceModel source, Models.Simulation.DeviceModelOverride overrideInfo);
     }
 
     public class DeviceModels : IDeviceModels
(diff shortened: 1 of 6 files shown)


SimulationAgent/DeviceTelemetry/DeviceTelemetryActor.cs
```diff
@@ -163,28 +163,31 @@ public void HandleEvent(ActorEvents e)
...
 
             var now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
-            if (now < this.whenToRun) return;
+            if (now < this.whenToRun) return null;
 
             switch (this.status)
             {
                 case ActorStatus.ReadyToStart:
-                    if (!this.deviceConnectionActor.Connected) return;
+                    if (!this.deviceConnectionActor.Connected) return null;
```
timlaverty: return "device not connected"

Services/DeviceModels.cs
```diff
@@ -78,6 +79,89 @@ public DeviceModel Get(string id)
             throw new ResourceNotFoundException();
         }
 
+        public DeviceModel OverrideDeviceModel(DeviceModel source, Models.Simulation.DeviceModelOverride overrideInfo)
```
timlaverty: do we have a github issue tracking revisiting the design of the override classes?  The more I think of it the more I think they're going to get unwieldy & hard to maintain in the future.

SimulationAgent/DeviceTelemetry/DeviceTelemetryActor.cs
dluc: 👍

Services/DeviceModels.cs
dluc: https://github.com/Azure/device-simulation-dotnet/issues/144

PR #143, opened 2017-11-30
Improved Azure IotHub Connection String Validation 

jillcary:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [x] Enhancement

Adds additional validation to Azure IotHub Connection Strings provided by the client.

Currently, if the user provides a connection string that does not have the correct read/write or service permissions, the service will silently fail. The new validation accounts for this and returns an error to the client if the string cannot be used to retrieve a list of devices and create/delete a device.

Also, if the service was restarted and the connection string was no longer locally stored, a null reference exception was thrown. This fix will check for null values before validation continues and return a meaningful error to the client. 

**Checklist:**

- [x] All tests passed


Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -131,19 +141,28 @@ public void ValidateConnectionString(string connectionString)
                 throw new InvalidIotHubConnectionStringFormatException(message);
             }
 
+            // check if KeyName permissions are either "iothubowner" or "service"
+            if (!(match.Groups[CONNSTRING_REGEX_KEYNAME].Value == OWNER_KEYNAME ||
+                  match.Groups[CONNSTRING_REGEX_KEYNAME].Value == SERVICE_KEYNAME))
```
dluc: as discussed offline, we cannot use this kind of validation, because users can create custom connection strings with custom keynames

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -118,7 +120,7 @@ public static SimulationApiModel FromServiceModel(Simulation value)
             return result;
         }
 
-        public void ValidateInputRequest(ILogger log)
+        public async Task ValidateInputRequest(ILogger log)
```
dluc: add "Async" suffix

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,10 +123,17 @@ public string RedactAndStore(string connectionString)
         /// value for the key, also checks if a connection to the IotHub
         /// can be made.
         /// </summary>
-        public void ValidateConnectionString(string connectionString)
+        public static async Task ValidateConnectionString(string connectionString)
```
dluc: async suffix, no static

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -165,6 +167,8 @@ public void ValidateInputRequest(ILogger log)
                 log.Error(INVALID_DATE, () => new { simulation = this });
                 throw new BadRequestException(INVALID_DATE, e);
             }
+
+            await IotHubConnectionStringManager.ValidateConnectionString(this.IotHub.ConnectionString);
```
dluc: avoid static calls, inject the IotHubConnectionStringManager similarly to how the logger is passed in

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -15,7 +17,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.IotHub
     public interface IIotHubConnectionStringManager
     {
         string GetIotHubConnectionString();
-        string RedactAndStore(string connectionString);
+        Task<string> RedactAndStore(string connectionString);
```
dluc: "Async" suffix

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -26,6 +28,9 @@ public class IotHubConnectionStringManager : IIotHubConnectionStringManager
         private const string CONNSTRING_REGEX_KEY = "key";
         private const string CONNSTRING_FILE_NAME = "custom_iothub_key.txt";
 
+        private const string OWNER_KEYNAME = "iothubowner";
+        private const string SERVICE_KEYNAME = "service";
```
dluc: these 2 constants are not needed

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -134,16 +146,35 @@ public void ValidateConnectionString(string connectionString)
...
+                    throw new IotHubConnectionException(message, e);
+                }
+
+                try
+                {
+                    // check read permissions
+                    await registryManager.GetDevicesAsync(1, CancellationToken.None);
+
+                    // check write permissions
+                    var testDevice = new Device();
```
dluc: what's the ID of the test device? I would give it an ID to explain what it is, in case the user sees it in the portal (note that RemoveDeviceAsync could fail and leave the device in hub).  Also, you might get a conflict if the device exists already, so I would add a suffix like "current time in milliseconds". Also consider that if RemoveDeviceAsync keeps failing, the hub will be polluted with test devices.

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,10 +123,17 @@ public string RedactAndStore(string connectionString)
...
+            // valid if default or null
+            if (connectionString == ServicesConfig.USE_DEFAULT_IOTHUB ||
+                connectionString.IsNullOrWhiteSpace())
+            {
+                return;
+            }
 
+            // check format of provided string
+            var match = Regex.Match(connectionString, CONNSTRING_REGEX);
             if (!match.Success)
```
dluc: consider splitting the method, e.g. each try/catch separately

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -134,16 +146,35 @@ public void ValidateConnectionString(string connectionString)
...
+                    var testDevice = new Device();
+                    testDevice = await registryManager.AddDeviceAsync(testDevice);
+                    await registryManager.RemoveDeviceAsync(testDevice);
+                }
+                catch (Exception e)
+                {
+                    string message = "Could not read or create devices with IotHub connection" +
+                                     "string provided. Check that the policy for the key allows " + 
+                                     "`Registry Read/Write` and `Service Connect` permissions.";
+                    throw new IotHubConnectionException(message, e);
```
dluc: nit: rename the exception to something like "IoTHubConnectionTestException" (or "validation" or "connection string validation" etc.)

Services.Test/IotHubConnectionStringManagerTest.cs
```diff
@@ -24,10 +24,10 @@ public IotHubConnectionStringManagerTest()
         }
 
         [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
-        public void ItThrowsOnInvalidConnStringFormat()
+        public async System.Threading.Tasks.Task ItThrowsOnInvalidConnStringFormat()
         {
             // Assert
-            Assert.Throws<InvalidIotHubConnectionStringFormatException>(() => this.target.RedactAndStore("foobar"));
+            await Assert.ThrowsAsync<InvalidIotHubConnectionStringFormatException>(() => this.target.RedactAndStore("foobar"));
```
timlaverty: Why do you need to await the assert?

Services/IotHub/IotHubConnectionStringManager.cs
jillcary: See updated methods below.

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
         /// value for the key, also checks if a connection to the IotHub
         /// can be made.
         /// </summary>
-        public void ValidateConnectionString(string connectionString)
+        public async Task ValidateConnectionStringAsync(string connectionString)
         {
-            var match = Regex.Match(connectionString, CONNSTRING_REGEX);
+            // valid if default or null
+            if (connectionString == ServicesConfig.USE_DEFAULT_IOTHUB ||
```
dluc: nit: make the comparison case insensitive

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
+
+            this.log.Debug("IotHub connection string provided is valid.", () => { });
+        }
+
+        /// <summary> Throws if unable to create a registry manager with a valid IotHub. </summary>
+        private void ValidateExistingIotHub(string connectionString)
+        {
+            try
+            {
+                var rm = RegistryManager.CreateFromConnectionString(connectionString);
```
dluc: nit: `rm` is not used

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
+            }
+            catch (Exception e)
+            {
+                string message = "Could not connect to IotHub with the connection " +
+                                 "string provided. Check that the key is valid and " +
+                                 "that the hub exists.";
+                this.log.Debug(message, () => { });
```
dluc: loglevel perhaps Warn/Error?

dluc: log the exception too

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
+            try
+            {
+                await registryManager.GetDevicesAsync(1, CancellationToken.None);
+            }
+            catch (Exception e)
+            {
+                string message = "Could not read devices with IotHub connection " +
+                                 "string provided. Check that the policy for the key allows " +
+                                 "`Registry Read/Write` and `Service Connect` permissions.";
+                this.log.Debug(message, () => { });
```
dluc: ditto, loglevel and log `e`

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
+                throw new IotHubConnectionException(message, e);
+            }
+
+        }
+
+        private async Task ValidateWritePermissionsAsync(string connectionString)
+        {
+            var registryManager = RegistryManager.CreateFromConnectionString(connectionString);
+
+            string testDeviceId = "test-device-creation" + DateTime.UtcNow.ToFileTimeUtc();
```
dluc: nit: make sure the suffix includes milliseconds

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
+            try
+            {
+                await registryManager.AddDeviceAsync(device);
+            }
+            catch (Exception e)
+            {
+                string message = "Could not create devices with IotHub connection " +
+                                 "string provided. Check that the policy for the key allows " +
+                                 "`Registry Read/Write` and `Service Connect` permissions.";
+                this.log.Debug(message, () => { });
```
dluc: loglevel and log `e`

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
                 }
-                catch (IOException)
+                catch (Exception e)
                 {
-                    string message = "Could not connect to IotHub with the connection" +
-                                     "string provided. Check that the key is valid and " +
-                                     "that the hub exists.";
-                    throw new IotHubConnectionException(message);
+                    string message = "Could not delete test device from IotHub. Device Id: " + testDeviceId;
+                    this.log.Error(message, () => new { testDeviceId });
```
dluc: log also the exception

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
+                    await registryManager.RemoveDeviceAsync(testDeviceId);
                 }
-                catch (IOException)
+                catch (Exception e)
                 {
-                    string message = "Could not connect to IotHub with the connection" +
-                                     "string provided. Check that the key is valid and " +
-                                     "that the hub exists.";
-                    throw new IotHubConnectionException(message);
+                    string message = "Could not delete test device from IotHub. Device Id: " + testDeviceId;
```
dluc: nit: no need to append the device id to the message

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
+                throw new IotHubConnectionException(message, e);
+            }
+
+            // Delete the test device that was created.
+            // If test device deletion fails, retry. Throw if unsuccessful.
+            int deleteRetryCount = 3;
+            Device response;
+            do
+            {
+                deleteRetryCount--;
```
dluc: nit: decrease this in line 234

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,122 @@ public string RedactAndStore(string connectionString)
...
                 }
+
+                response = await registryManager.GetDeviceAsync(testDeviceId);
+
+            } while (response != null && deleteRetryCount > 0);
+
+            if (response != null)
+            {
+                string message = "Could not delete test device from IotHub. Device Id: " + testDeviceId;
+                this.log.Error(message, () => new { testDeviceId });
```
dluc: log also the number of attempts.  No need to include the device id in the msg (avoid string manipulation and reduce GC work)

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -118,33 +122,128 @@ public string RedactAndStore(string connectionString)
...
+            catch (Exception e)
+            {
+                string message = "Could not create devices with IotHub connection " +
+                                 "string provided. Check that the policy for the key allows " +
+                                 "`Registry Read/Write` and `Service Connect` permissions.";
+                this.log.Error(message, () => new { e });
+                throw new IotHubConnectionException(message, e);
+            }
+
+            // Delete the test device that was created.
```
timlaverty: consider refactoring to separate function

PR #146, opened 2017-12-01
Internal scripts for increasing/decreasing/random

dluc:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

New logic for customizing sensors and selected pre-configured behavior like increasing, decreasing and random numbers.

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/146)

### DIFF
--- a/SimulationAgent/SimulationRunner.cs
+++ b/SimulationAgent/SimulationRunner.cs
@@ -28,6 +28,9 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent
         // ID prefix of the simulated devices, used with Azure IoT Hub
         private const string DEVICE_ID_PREFIX = "Simulated.";
 
+        // ID used for custom device models, where the list of sensors is provided by the user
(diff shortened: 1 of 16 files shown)


SimulationAgent/SimulationRunner.cs
```diff
@@ -28,6 +28,9 @@ public class SimulationRunner : ISimulationRunner
         // ID prefix of the simulated devices, used with Azure IoT Hub
         private const string DEVICE_ID_PREFIX = "Simulated.";
 
+        // ID used for custom device models, where the list of sensors is provided by the user
+        private const string CUSTOM_DEVICE_MODEL_ID = "custom";
```
jillcary: Should this also be capitalized?

dluc: the code assume it's a lowercase value, see the where the value is used. The value sent from the UI is lowercased, to make it case insensitive.

PR #153, opened 2017-12-14
When available, return a URL to the metrics of the pre-provisioned hub

dluc:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Add new configuration settings storing information about the IoT Hub deployed.  If the information is available, the status endpoint now returns a new property, with a URL to open the hub metrics in the Azure portal.

**Checklist:**

- [x] All tests passed

### DIFF
--- a/WebService/Runtime/Config.cs
+++ b/WebService/Runtime/Config.cs
@@ -28,6 +28,9 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.Runtime
 
         // Simulation speed configuration
         IRateLimitingConfig RateLimitingConfig { get; }
+
+        // Deployment details
+        IDeploymentConfig DeploymentConfig { get; }
     }
 
     public class Config : IConfig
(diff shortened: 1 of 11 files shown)


WebService/Runtime/Config.cs
```diff
@@ -87,6 +96,18 @@ public Config(IConfigData configData)
...
             this.ClientAuthConfig = GetClientAuthConfig(configData);
             this.RateLimitingConfig = GetRateLimitingConfig(configData);
+            this.DeploymentConfig = GetDeploymentConfig(configData);
+        }
+
+        private static IDeploymentConfig GetDeploymentConfig(IConfigData configData)
+        {
+            return new DeploymentConfig
+            {
+                AzureSubscriptionDomain = configData.GetString(AZURE_SUBSCRIPTION_DOMAIN, "undefined.onmicrosoft.com"),
```
jillcary: Is this different just to show it's a domain or is it a required default value?

WebService/appsettings.ini
```diff
@@ -12,6 +12,25 @@ iothub_data_folder = ./data/iothub/
...
+# The value is used to create a URL taking to the IoT Hub metrics in the Azure portal.
+azure_subscription_id = "${?PCS_SUBSCRIPTION_ID}"
+# Azure resource group where the Azure IoT Hub is deployed, e.g. "abcd1234-5678-1234-abcd-abcd5678abcd".
+# The value is optional because the service can be deployed without a hub.
+# The value is used to create a URL taking to the IoT Hub metrics in the Azure portal.
+azure_resource_group = "${?PCS_RESOURCE_GROUP}"
+# IoT Hub name, e.g. "mytest3507e89".
+# The value is optional because the service can be deployed without a hub.
+# The value is used to create a URL taking to the IoT Hub metrics in the Azure portal.
+azure_iothub_name = "${?PCS_IOHUB_NAME}"
```
jillcary: Is this value required? The hub name can be extracted from the connection string.

WebService/Runtime/Config.cs
dluc: not a required value, just a default non-valid one

WebService/appsettings.ini
dluc: I think it might be different from the one used in the connection string, e.g. in cases where the name is long

PR #154, opened 2017-12-18
Create devices using batch operations

dluc:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Use batch operations to create the simulated devices.
Delete devices when a simulation is deleted (not when it's stopped/completed).

**Checklist:**

- [x] All tests passed

### DIFF
--- a/Services/DeviceModels.cs
+++ b/Services/DeviceModels.cs
@@ -22,6 +22,9 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
 
     public class DeviceModels : IDeviceModels
     {
+        // ID used for custom device models, where the list of sensors is provided by the user
+        public const string CUSTOM_DEVICE_MODEL_ID = "custom";
+
         private const string EXT = ".json";
 
         private readonly IServicesConfig config;
--- a/Services/Devices.cs
+++ b/Services/Devices.cs
@@ -279,9 +279,6 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         public void SetCurrentIotHub()
         {
             string connString = this.connectionStringManager.GetIotHubConnectionString();
-
-            connString = "HostName=devissim003useast9dd22.azure-devices.net;SharedAccessKeyName=iothubowner;SharedAccessKey=ymxUjxVhTfqVLaT1SmZF7oETlWdTUyOR6cBE8npaGD0=";
-
             this.registry = RegistryManager.CreateFromConnectionString(connString);
             this.ioTHubHostName = IotHubConnectionStringBuilder.Create(connString).HostName;
             this.log.Info("Selected active IoT Hub for devices", () => new { this.ioTHubHostName });
--- a/Services/Devices.cs
+++ b/Services/Devices.cs
@@ -21,6 +21,11 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
     public interface IDevices
     {
         /// <summary>
+        /// Set the current IoT Hub using either the user provided one or the configuration settings
+        /// </summary>
+        void SetCurrentIotHub();
+
+        /// <summary>
         /// Get a client for the device
         /// </summary>
         IDeviceClient GetClient(Device device, IoTHubProtocol protocol);
@@ -49,11 +54,6 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         /// Delete a list of devices
         /// </summary>
         Task DeleteListAsync(IEnumerable<string> deviceIds);
-
-        /// <summary> 
-        /// Set the current IoT Hub using either the user provided one or the configuration settings 
-        /// </summary>
-        void SetCurrentIotHub();
 
         /// <summary>
         /// Generate a device Id
@@ -93,6 +93,17 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
             this.twinReadsWritesEnabled = config.TwinReadWriteEnabled;
             this.registryCount = -1;
             this.setupDone = false;
+        }
+
+        /// <summary>
+        /// Get IoTHub connection string from either the user provided value or the configuration
+        /// </summary>
+        public void SetCurrentIotHub()
+        {
+            string connString = this.connectionStringManager.GetIotHubConnectionString();
+            this.registry = RegistryManager.CreateFromConnectionString(connString);
+            this.ioTHubHostName = IotHubConnectionStringBuilder.Create(connString).HostName;
+            this.log.Info("Selected active IoT Hub for devices", () => new { this.ioTHubHostName });
         }
 
         /// <summary>
@@ -271,17 +282,6 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
                 this.log.Error("Failed to delete devices", () => new { error });
                 throw;
             }
-        }
-
-        /// <summary> 
-        /// Get IoTHub connection string from either the user provided value or the configuration 
-        /// </summary>
-        public void SetCurrentIotHub()
-        {
-            string connString = this.connectionStringManager.GetIotHubConnectionString();
-            this.registry = RegistryManager.CreateFromConnectionString(connString);
-            this.ioTHubHostName = IotHubConnectionStringBuilder.Create(connString).HostName;
-            this.log.Info("Selected active IoT Hub for devices", () => new { this.ioTHubHostName });
         }
 
         /// <summary>
--- a/Services/Devices.cs
+++ b/Services/Devices.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         // the simulation. When the simulation is running the registry is not used anymore.
         private const uint REGISTRY_LIMIT_REQUESTS = 1000;
 
-        // When working with batches, this is the max size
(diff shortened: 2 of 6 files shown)


Services/Devices.cs
```diff
@@ -174,17 +199,102 @@ public IDeviceClient GetClient(Device device, IoTHubProtocol protocol)
...
+        }
+
         /// <summary> 
         /// Get IoTHub connection string from either the user provided value or the configuration 
         /// </summary>
         public void SetCurrentIotHub()
         {
             string connString = this.connectionStringManager.GetIotHubConnectionString();
+
+            connString = "HostName=devissim003useast9dd22.azure-devices.net;SharedAccessKeyName=iothubowner;SharedAccessKey=ymxUjxVhTfqVLaT1SmZF7oETlWdTUyOR6cBE8npaGD0=";
```
timlaverty: this looks like an actual key; I suspect you didn't mean to include this in your PR.

dluc:
@timlaverty in the last tests I ran, when creating a simulation for the first time, with more than ~600 devices, the service fails to create all the devices. There are 2-3 places where we need a better retry logic but I didn't have time to investigate.  This change mitigates that problem, and considerably speeds up the simulation start.  We will still need to fix the retry problems, but in the meantime I'd like to have this in place.

Services/Devices.cs
```diff
@@ -174,15 +210,86 @@ public IDeviceClient GetClient(Device device, IoTHubProtocol protocol)
...
+                () => new { Count = deviceIds.Count(), Batches = batches.Length, REGISTRY_MAX_BATCH_SIZE });
+
+            for (var batchNumber = 0; batchNumber < batches.Length; batchNumber++)
+            {
+                var batch = batches[batchNumber];
+
+                this.log.Info("Creating devices batch",
+                    () => new { batchNumber, batchSize = batch.Count() });
+
+                BulkRegistryOperationResult result = await this.registry.AddDevices2Async(
```
jillcary: What happens if creation fails? Do we stop the simulation?

Services/Simulations.cs
```diff
@@ -124,6 +165,7 @@ public class Simulations : ISimulations
         }
 
         /// <summary>
+        /// Create or Replace a simulation
```
jillcary: Nit. Summary content for Upsert below has punctuation. "Upsert the simulation" may also be redundant now with the added summary content.

Services/Simulations.cs
```diff
@@ -228,9 +273,37 @@ public class Simulations : ISimulations
...
         }
 
+        /// <summary>
+        /// Delete a simulation and its devices
+        /// </summary>
         public async Task DeleteAsync(string id)
         {
+            // Delete devices first
+            var deviceIds = this.GetDeviceIds(await this.GetAsync(id));
+            await this.devices.DeleteListAsync(deviceIds);
```
jillcary: If this fails do the devices stay in the IotHub?

SimulationAgent/SimulationRunner.cs
```diff
@@ -127,6 +126,24 @@ public void Start(Services.Models.Simulation simulation)
...
+
+                    // This will ignore existing devices, creating only the missing ones
+                    this.devices.CreateListAsync(devices).Wait(TimeSpan.FromSeconds(30));
+                }
+                catch (Exception e)
+                {
+                    this.running = false;
+                    this.starting = false;
+                    this.log.Error("Failed to create devices", () => new { e });
+
```
jillcary: If creation fails, do we delete the devices that may have been created on the hub?

Services/DeviceModels.cs
```diff
@@ -22,6 +22,9 @@ public interface IDeviceModels
 
     public class DeviceModels : IDeviceModels
     {
+        // ID used for custom device models, where the list of sensors is provided by the user
+        public const string CUSTOM_DEVICE_MODEL_ID = "custom";
```
timlaverty: out of scope?  I like the change though- makes it more readable

Services/Devices.cs
```diff
@@ -49,6 +68,12 @@ public class Devices : IDevices
         // the simulation. When the simulation is running the registry is not used anymore.
         private const uint REGISTRY_LIMIT_REQUESTS = 1000;
 
+        // When working with batches, this is the max size
```
timlaverty: super nit: "..., this is the max size that the batch insert and delete APIs allow"

Services/Devices.cs
```diff
@@ -174,15 +210,86 @@ public IDeviceClient GetClient(Device device, IoTHubProtocol protocol)
...
+            {
+                var batch = batches[batchNumber];
+
+                this.log.Info("Creating devices batch",
+                    () => new { batchNumber, batchSize = batch.Count() });
+
+                BulkRegistryOperationResult result = await this.registry.AddDevices2Async(
+                    batch.Select(id => new Azure.Devices.Device(id)), CancellationToken.None);
+
+                this.log.Info("Devices batch created",
```
timlaverty: nit: log the batch id so you can better track what failed?

Services/Devices.cs
timlaverty: What happens if the hub isn't available or there is something wrong w/t he iterator of devices you pass?  Does this api throw or will it always succeed and return result?

Services/Devices.cs
```diff
@@ -265,5 +372,14 @@ private Azure.Devices.Client.DeviceClient GetDeviceSdkClient(Device device, IoTH
 
             return sdkClient;
         }
+
+        private IEnumerable<IEnumerable<T>> SplitArray<T>(IReadOnlyCollection<T> array, int size)
+        {
+            var count = (int) Math.Ceiling((float) array.Count / size);
```
timlaverty: Unfortunate that there's no built in array function that allows splitting of arrays into sub-arrays.  I looked around & couldn't find a better way to do this.  did you consider linq?  something like:

arrayString.GroupBy(s => i++/lengthToSplit).Select(g => g.ToArray()).ToList()

Services/Simulations.cs
timlaverty: I'd be ok if they did as delete isn't used by anyone but us from postman right now.

SimulationAgent/SimulationRunner.cs
```diff
@@ -127,6 +126,24 @@ public void Start(Services.Models.Simulation simulation)
                 // the active hub, e.g. in case the user provided a custom connection string.
                 this.devices.SetCurrentIotHub();
 
+                // Create the devices
+                try
+                {
+                    var devices = this.simulations.GetDeviceIds(simulation);
+
+                    // This will ignore existing devices, creating only the missing ones
+                    this.devices.CreateListAsync(devices).Wait(TimeSpan.FromSeconds(30));
```
timlaverty: why 30s?  Is there a chance you'll end up in the "real world" and need more time?  

Also, nit: move to constant?

Services/DeviceModels.cs
dluc: it was in scope at some point, to avoid duplicating the constant which was used here and in `SimulationRunner.cs`.  Eventually due to refactoring it's not strictly required, but I think the constant belongs here rather than in the runner.

Services/Devices.cs
dluc: 👍

Services/Devices.cs
dluc: 👍 also on line 267

Services/Devices.cs
dluc: exceptions will occur in case of timeout, wrong conn string, etc. The exceptions will bubble up to the caller, in this case `SimulationRunner.Start`, which is designed to handle them with a retry:

```csharp
try
{
    var devices = this.simulations.GetDeviceIds(simulation);

    // This will ignore existing devices, creating only the missing ones
    this.devices.CreateListAsync(devices).Wait(TimeSpan.FromSeconds(30));
}
catch (Exception e)
{
    this.running = false;
    this.starting = false;
    this.log.Error("Failed to create devices", () => new { e });

    // Return and retry
    return;
}
```

dluc: If any of the devices passed already exists (which is fine), the operation will succeed, tracking a 409 (conflict) in the `result` variable.

Services/Devices.cs
dluc: nice, I like linq. I find that "group by" odd to read though. Future-wise we'll have to do a lot of partitioning, e.g. to run across multiple instances, so I'm sure we'll revisit this code

Services/Simulations.cs
dluc: 👍

Services/Simulations.cs
dluc: @jillcary yes, there is no error handling/retry here, because we use this endpoint only for tests/debugging right now. In future we'll probably promote this to be the official behavior, at which point we'll want to handle failures as you suggest

SimulationAgent/SimulationRunner.cs
dluc: Moved to constant.
The creation of 1000 devices takes about 2-3 seconds. If it takes longer, this will trigger a timeout exception and the agent will retry. When we go big, we'll refactor/improve this to scale out horizontally.

SimulationAgent/SimulationRunner.cs
dluc: Not at this stage, we only care that devices exists. (this was the case also before, also in RM)

PR #155, opened 2017-12-19
Throw error on all-whitespace connection string & validate default hub

jillcary:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

When the user entered a custom connection string that was all whitespace, the backend would read it as null and set it to the default hub. If the solution didn't have a pre-provisioned hub, the service would run and silently fail. 

This change:
Adds additional validation for all-whitespace strings
Checks that the default hub can be connected to and has the proper permissions. 

**Checklist:**

- [x] All tests passed

### DIFF
--- a/Services/IotHub/IotHubConnectionStringManager.cs
+++ b/Services/IotHub/IotHubConnectionStringManager.cs
@@ -159,17 +159,13 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.IotHub
         /// </summary>
         private bool IsDefaultHub(string connectionString)
         {
-            if (connectionString == null ||
+            return
+                connectionString == null ||
                 connectionString == string.Empty ||
                 string.Equals(
                     connectionString.Trim(),
                     ServicesConfig.USE_DEFAULT_IOTHUB,
-                    StringComparison.OrdinalIgnoreCase))
-            {
-                return true;
-            }
-
-            return false;
+                    StringComparison.OrdinalIgnoreCase);
         }
 
         /// <summary> Throws if unable to create a registry manager with a valid IotHub. </summary>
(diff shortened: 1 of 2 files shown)


Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -56,7 +56,7 @@ public string GetIotHubConnectionString()
             string customIotHub = this.ReadFromFile();
 
             // if no user provided hub is stored, use the pre-provisioned hub
```
dluc: nit: could you rephrase this comment, hard to read

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -56,7 +56,7 @@ public string GetIotHubConnectionString()
             string customIotHub = this.ReadFromFile();
 
             // if no user provided hub is stored, use the pre-provisioned hub 
-            if (customIotHub.IsNullOrWhiteSpace())
+            if (this.IsDefaultHub(customIotHub))
             {
                 this.log.Debug("Using IotHub connection string stored in config.", () => { });
```
dluc: please change this to Info level

WebService/v1/Models/SimulationApiModel/SimulationIotHub.cs
```diff
@@ -27,9 +29,24 @@ public SimulationIotHub(string connectionString) : this()
...
+                connectionString == string.Empty ||
+                string.Equals(
+                    connectionString.Trim(),
+                    USE_DEFAULT_IOTHUB,
+                    StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            return false;
```
dluc: merge the return into one statement

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -157,6 +153,25 @@ public string GetIotHubConnectionString()
...
+                connectionString == string.Empty ||
+                string.Equals(
+                    connectionString.Trim(),
+                    ServicesConfig.USE_DEFAULT_IOTHUB,
+                    StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            return false;
```
dluc: merge into one statement: return (....)

PR #159, opened 2018-01-22
Improved gps location device simulation

joelhulen:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [x] Enhancement

The previous version never actually changed the truck's location after the first pass. This version randomizes the changed distance, plus it adds a rounding function that handles the positive and negative values of the lat/long coordinates equally well. This results in an improved randomized route for the trucks. Not 100% realistic (data would need to be incorporated for that), but good enough for a simulation.

**Checklist:**

- [ ] All tests passed


dluc:
@joelhulen thanks for the PR, we're looking into it

PR #162, opened 2018-01-27
Remove old code used to detect simulation stop via cancellation token

dluc:
# Type of change?

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Remove old code used to stop the simulation via cancellation token. The old code was interfering with the new state machine, where timeouts are actually errors.

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/162)

### DIFF
--- a/Services/Devices.cs
+++ b/Services/Devices.cs
@@ -3,7 +3,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices;
 using Microsoft.Azure.Devices.Common.Exceptions;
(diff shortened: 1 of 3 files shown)


Services/Devices.cs
```diff
@@ -183,13 +184,13 @@ public IDeviceClient GetClient(Device device, IoTHubProtocol protocol)
             {
                 if (e.InnerException != null && e.InnerException.GetType() == typeof(TaskCanceledException))
                 {
-                    // We get here when the cancellation token is triggered, which is fine
-                    this.log.Debug("Device creation task canceled", () => new { deviceId, e.Message });
-                    return null;
+                    var timeSpent = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - now.ToUnixTimeMilliseconds();
+                    this.log.Error("Create device task timed out", () => new { timeSpent, deviceId, e.Message });
+                    throw;
```
timlaverty: What was happening before when we didn't throw?  Did the device just sit in an indeterminate state?  ...or did returning null cause something outside of here to break?

dluc: The `null` returned here was used as a valid device object in one of the state machines, leading to nullref exceptions later on, and bringing that SM in a non-recoverable state.

ppathan:
I was talking about these names, agree that this are from SDKs
AddDevices2Async(IEnumerable<Device> devices);
RemoveDevices2Async(IEnumerable<Device> devices, bool forceRemove);

PR #173, opened 2018-03-01
Add Device Properties logic to internal simulation

jillcary:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [x] Enhancement

### Context:
This change adds back the functionality for the internal simulation to update the internal device properties state. This also adds some new classes for the SimulationAgent to manage device properties. 

### This change includes:
1. Remove commented out code for Device Twin management.
1. Addition of the **InternalDeviceState** Class
    * The **"InternalDeviceState"** has been modified to encompass the following two entities:
        1. The **"DeviceSimulationState"** refers to sensor data that is simulated by the DeviceStateActor (ex: temperature, humidity...)
        1. The **"DeviceProperties"** refers to the values that are to be written to the IoT Hub Device Twin (ex: firmware, device type...)
    * **Callbacks** -  This InternalDeviceState is then passed to the Javascript Interpreter so the scripts can update both the device state and the properties via a callback. 
    * **Change Notification** The InternalDeviceState also writes custom getters and setters so that modification of the properties sets a flag indicating the values have changed and should be updated on the IoT Hub. 
1. Pass the InternalDeviceState object instead of a ```Dictionary<string, object>``` to the ScriptInterpreter
1. Add restore device property state logic to device scripts
1. Add Stubbed DevicePropertyActor (functionality coming in future PR 

### Notes & Assumptions:
* **Language and nomenclature** - "Device Properties" are used instead of "Reported Properties" or "Device Twin" for the internal device state. This is because this is the language that is to be used in the Device Model.
* It is possible we could use a ```ConcurrentDictionary``` for the InternalDeviceState class as well

### Out of scope for this PR (for future PR):
- sending device properties to IoT Hub

**Checklist:**

- [x] All tests passed


Services/Models/DeviceTwin.cs
```diff
@@ -17,9 +16,9 @@ public class DeviceTwin
         public string ETag { get; set; }
         public string DeviceId { get; set; }
         public bool IsSimulated { get; set; }
-        public Dictionary<string, JToken> DesiredProperties { get; set; }
-        public Dictionary<string, JToken> ReportedProperties { get; set; }
-        public Dictionary<string, JToken> Tags { get; set; }
+        public Dictionary<string, object> DesiredProperties { get; set; }
+        public Dictionary<string, object> ReportedProperties { get; set; }
+        public Dictionary<string, object> Tags { get; set; }
```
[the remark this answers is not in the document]
jillcary: Reverting changes, as this class is not needed for the internal twin management.

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+        bool PropertyChanged { get; set; }
+        Dictionary<string, object> GetProperties();
+        bool HasProperty(string key);
+        object GetProperty(string key);
+        void SetProperty(string key, object value);
+        Dictionary<string, object> GetState();
+        void SetState(Dictionary<string, object> state);
+        bool HasStateValue(string key);
+        object GetStateValue(string key);
+        void SetStateValue(string key, object value);
```
dluc: could you revisit these names consistency? you could also consider reusing one internal class for properties and state, e.g.
```
IInternalDeviceState->Properties->GetAll()
IInternalDeviceState->Properties->Set(k, v)
IInternalDeviceState->Properties->Get(k)
IInternalDeviceState->Properties->Has(k)
IInternalDeviceState->Properties->Changed(k)
IInternalDeviceState->State->GetAll()
IInternalDeviceState->State->Set(k, v)
IInternalDeviceState->State->Get(k)
IInternalDeviceState->State->Has(k)
IInternalDeviceState->State->Changed(k)
```

Services/data/devicemodels/scripts/chiller-01-state.js
```diff
@@ -33,11 +39,29 @@ function restoreState(previousState) {
...
+}
+
 /**
  * Simple formula generating a random value around the average
  * in between min and max
+ *
+ * @returns random value with given parameters
  */
 function vary(avg, percentage, min, max) {
+
```
maple10001: Are we adding a new line under each function signatures on purpose?

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+    /// </summary>
+    public class InternalDeviceState : IInternalDeviceState
+    {
+        public const string CALC_TELEMETRY = "CalculateRandomizedTelemetry";
+
+        private ILogger log;
+
+        // TODO https://github.com/Azure/device-simulation-dotnet/issues/171
+        // consider using thread safe concurrent dictionary here in case multiple scripts
+        // try to modify the state at the same time.
```
dluc: why not use a concurrent dictionary now?

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+
+        private ILogger log;
+
+        // TODO https://github.com/Azure/device-simulation-dotnet/issues/171
+        // consider using thread safe concurrent dictionary here in case multiple scripts
+        // try to modify the state at the same time. 
+
+        // The virtual state of the simulated device. The simulationState is
+        // periodically updated using an external script.
+        private Dictionary<string, object> simulationState;
```
dluc: FYI, this name confuses me a bit, e.g. is "simulation state" something else than "internal device state" ?

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+
+        // The virtual state of the simulated device properties. The reported properties,
+        // if changed, will be written to the IoT Hub.
```
dluc: I would reduce the terms count, e.g. is "virtual state" something different from "internal state" and "simulation state" ? also, I think we don't need to mention the "if changed" bit here

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+        private Dictionary<string, object>  properties;
+
+        // flag that indicates when the twin needs to be pushed to the IoT Hub
```
dluc: ditto: na new term "twin" :-) which makes me wonder if that's something different from "internal state" "virtual state" etc.

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+                else
+                {
+                    this.simulationState.Add(key, value);
+                }
+            }
+        }
+
+        private Dictionary<string, object> SetupTelemetry(DeviceModel deviceModel)
+        {
+            // put telemetry properties in state
```
dluc: note, the state does not contain "only" telemetry data, also simulation state, e.g. see the "simulation_state" field here https://github.com/Azure/device-simulation-dotnet/blob/azure-iot-pcs-simulation/Services/data/devicemodels/chiller-01.json

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+
+            // Ensure the state contains the "online" key
+            if (!state.ContainsKey("online"))
+            {
+                state["online"] = true;
+            }
+
+            // TODO:This is used to control whether telemetry is calculated in UpdateDeviceState.
+            // methods can turn telemetry off/on; e.g. setting temp high- turnoff, set low, turn on
+            // it would be better to do this at the telemetry item level - we should add this in the future
```
dluc: could you add a Github issue tracking this please?

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+
+            return state;
+        }
+
+        /// <summary>
+        /// Initializes device twin with initial device twin properties from the device model.
+        /// </summary>
+        private Dictionary<string, object> SetupProperties(DeviceModel deviceModel)
+        {
+
```
dluc: empty line

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+            Dictionary<string, object> result = new Dictionary<string, object>();
+
+            // add device properties to the properties dictionary. These properties will be written
```
dluc: hypernit: please start comments with capital letter. About the rest of the comment, I'm not sure if we should call out the simulation behavior in here. Anyhow, I would partially rephrase saying "These properties will be stored as 'reported properties' in the 'device twin' in Azure IoT Hub"

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+            
+            this.log = log;
+        }
+
+        /// <summary>
+        /// Returns true if key exists
+        /// </summary>
+        public bool HasProperty(string key)
+        {
+            // lock properties as mulitple scripts may try to access key at the same time.
```
dluc: typo

Services/Models/InternalDeviceState.cs
```diff
@@ -0,0 +1,247 @@
...
+                return this.properties[key];
+            }
+        }
+
+        /// <summary>
+        /// Retrieve all device properties from the DeviceTwin reported properties as a read only dictionary
+        /// </summary>
+        public Dictionary<string, object> GetProperties()
+        {
+            // lock properties as mulitple scripts may try to access key at the same time.
```
dluc: typo

Services/Devices.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.Azure.Devices;
```
jillcary: Will remove changes to this file.

Services/DeviceClient.cs
```diff
@@ -6,6 +6,7 @@
 using System.Threading.Tasks;
```
jillcary: Will also remove changes to this file.

Services/DeviceClient.cs
```diff
@@ -105,33 +108,10 @@ public class DeviceClient : IDeviceClient
             await this.SendRawMessageAsync(eventMessage);
         }
 
-        public Task UpdateTwinAsync(Device device)
+        public async Task UpdatePropertiesAsync(Device device)
```
dluc: not sure if I'm missing something here. There is an `async` in the method signature, but the method body doesn't have an `await`, does it compile?

Services/DeviceClient.cs
```diff
@@ -105,33 +108,10 @@ public class DeviceClient : IDeviceClient
...
-            }
-
-            // Write properties
-            var reportedProperties = DictionaryToTwinCollection(device.Twin.ReportedProperties);
-            await this.rateLimiting.LimitTwinWritesAsync(
-                () => this.client.UpdateReportedPropertiesAsync(reportedProperties));
-            */
-            return Task.CompletedTask;
+            // TODO see https://github.com/Azure/device-simulation-dotnet/tree/send-twin-updates
+            // for future PR
```
dluc: could you expand the comment, not sure what one will be looking for following that link

Services/DeviceClient.cs
dluc: do you think I should wait for the final version before completing the CR?

Services/Devices.cs
dluc: I assume there is no need to depend on rateLimiting, right?

Services/DeviceClient.cs
jillcary: Updated -- removed changes. To be submitted with pr for https://github.com/Azure/device-simulation-dotnet/tree/send-twin-updates

Services/Simulation/InternalInterpreter.cs
```diff
@@ -47,43 +48,42 @@ public InternalInterpreter(ILogger logger)
             this.random = new Random();
         }
 
-        public Dictionary<string, object> Invoke(
+        public void Invoke(
```
ppathan: So looks like we don't need return values?

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -182,20 +196,42 @@ private void UpdateState(JsValue data)
 
             this.log.Debug("Updating state from the script", () => new { data, this.deviceState });
 
-            stateChanges = this.JsValueToDictionary((JsValue) data);
+            stateChanges = JsValueToDictionary((JsValue)data);
```
ppathan: nit: ```this``` Other places we are using this.

Services/data/devicemodels/scripts/chiller-01-state.js
```diff
@@ -33,9 +37,25 @@ function restoreState(previousState) {
...
 }
 
+/**
+ * Restore the global properties using data from the previous iteration.
+ *
+ * @param previousProperties The output of main() from the previous iteration
+ */
+function restoreProperties(previousProperties) {
+    // If the previous properties are null, force the default properties
+    if (previousProperties !== undefined && previousProperties !== null) {
```
ppathan: ``` if(previousProperties)``` should work for both.

Services/data/devicemodels/scripts/elevator-01-state.js
```diff
@@ -32,6 +37,20 @@ function restoreState(previousState) {
...
 }
 
+/**
+ * Restore the global properties using data from the previous iteration.
+ *
+ * @param previousProperties The output of main() from the previous iteration
+ */
+function restoreProperties(previousProperties) {
+    // If the previous properties are null, force the default properties
+    if (previousProperties !== undefined && previousProperties !== null) {
```
ppathan: ditto

Services/data/devicemodels/scripts/truck-01-state.js
```diff
@@ -33,6 +38,20 @@ function restoreState(previousState) {
...
 }
 
+/**
+ * Restore the global properties using data from the previous iteration.
+ *
+ * @param previousProperties The output of main() from the previous iteration
+ */
+function restoreProperties(previousProperties) {
+    // If the previous properties are null, force the default properties
+    if (previousProperties !== undefined && previousProperties !== null) {
```
ppathan: ditto

Services.Test/InternalDevicePropertiesTest.cs
```diff
@@ -0,0 +1,222 @@
...
+using Xunit.Abstractions;
+using static Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models.DeviceModel;
+
+namespace Services.Test
+{
+    public class InternalDevicePropertiesTest
+    {
+        private InternalDeviceProperties target;
+
+        public InternalDevicePropertiesTest(ITestOutputHelper log)
```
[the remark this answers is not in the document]
jillcary: Not used for anything in this test in particular. Used throughout the tests in Simulation so the tests can log if needed.

Services.Test/InternalDevicePropertiesTest.cs
```diff
@@ -0,0 +1,222 @@
...
+            // Act
+            var result = this.target.Get(key);
+
+            // Assert
+            Assert.NotNull(result);
+            Assert.Equal(result, value);
+        }
+
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void SetPropertyTest()
```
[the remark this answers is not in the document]
jillcary: There are two new classes added. Upon discussion, we decided it was best for properties not to inherit from DeviceState In case device state behavior changes in the future and breaks the properties logic for checking if a value is "dirty" and needs to be updated to the twin. 

So yes, they are the same, but it was intentional.

Services.Test/DevicesTest.cs
```diff
@@ -4,6 +4,7 @@
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices;
 using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Concurrency;
```
dluc: I don't see other changes in this file, can we revert this change?

Services/Diagnostics/ActorsLogger.cs
```diff
@@ -288,6 +294,48 @@ public void TelemetryFailed()
             this.LogTelemetry("FAILED");
         }
 
+        public void DeviceTwinUpdateScheduled(long time)
+        {
+            if (!this.enabled) return;
+
+            var msg = DateTimeOffset.FromUnixTimeMilliseconds(time).ToString(DATE_FORMAT);
```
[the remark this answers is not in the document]
dluc: the `time` value is passed as a parameter, and the value is usually in the future

Services/Diagnostics/ActorsLogger.cs
```diff
@@ -288,6 +294,48 @@ public void TelemetryFailed()
...
+            this.Log("Twin update retry scheduled at: " + msg);
+            this.LogTwin("Retry scheduled at: " + msg);
+        }
+
+        public void UpdatingDeviceTwin()
+        {
```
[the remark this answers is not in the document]
dluc: @jillcary are these methods supposed to be used now?

Services/Models/InternalDeviceProperties.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        public InternalDeviceProperties()
+        {
+            this.dictionary = new ConcurrentDictionary<string, object>();
+            this.changed = false;
+        }
+
+        public InternalDeviceProperties(DeviceModel deviceModel)
+        {
+            var intitalProperties = this.SetupProperties(deviceModel);
```
dluc: typo

dluc: to avoid code duplication and risk of diverging, you could refactor this method to extend the other ctor
```
public InternalDeviceProperties(DeviceModel deviceModel): this()
{
    this.dictionary = new ConcurrentDictionary<string, object>(
        this.SetupProperties(deviceModel));
}
```

Services/Models/InternalDeviceProperties.cs
```diff
@@ -0,0 +1,110 @@
...
+{
+    public interface IInternalDeviceProperties
+    {
+        Dictionary<string, object> GetAll();
+        void SetAll(Dictionary<string, object> newState);
+        bool Has(string key);
+        object Get(string key);
+        void Set(string key, object value);
+        bool Changed { get; }
+        void ResetChanged();
```
dluc: how come that all these methods are never used?

Services/Models/DeviceTwin.cs
```diff
@@ -34,35 +34,17 @@ public DeviceTwin(Twin twin)
...
-        JToken:     IJEnumerable<JToken>, IEnumerable<JToken>, IEnumerable, IJsonLineInfo, ICloneable, IDynamicMetaObjectProvider
-        */
         private static Dictionary<string, JToken> TwinCollectionToDictionary(TwinCollection x)
         {
             var result = new Dictionary<string, JToken>();
 
             if (x == null) return result;
 
-            foreach (KeyValuePair<string, object> twin in x)
+            foreach (KeyValuePair<string, JToken> twin in x)
```
dluc: As mentioned earlier, it's ok refactoring this class, however we need to test all the scenarios if we do so. We know that the current mapping function works. To avoid regressions we need tests.

Services.Test/DevicesTest.cs
dluc: just checked: this change can be reverted

Services.Test/InternalDevicePropertiesTest.cs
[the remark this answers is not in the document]
dluc: it's a convention to inject the test logger in all our tests, in case we need to log. Please leave it. It might be useful to add a comment explaining how to use it (across all our tests, in a separate PR)

Services.Test/InternalDevicePropertiesTest.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        private InternalDeviceProperties GetTestProperties()
+        {
+            return new InternalDeviceProperties(GetTestChillerModel());
+        }
+
+        /// <summary>
+        /// Returns the a test chiller model
+        /// </summary>
+        private DeviceModel GetTestChillerModel()
```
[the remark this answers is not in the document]
dluc: I understand it might be a matter of taste, but I'd suggest to avoid coupling data providers across tests. In my experience they increase the chance of failures when adding tests

Services.Test/Simulation/JavascriptInterpreterTest.cs
```diff
@@ -41,14 +44,16 @@ public JavascriptInterpreterTest(ITestOutputHelper log)
...
+
             var filename = "chiller-01-state.js";
             var context = new Dictionary<string, object>
             {
                 ["currentTime"] = DateTimeOffset.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:sszzz"),
                 ["deviceId"] = "device-123",
                 ["deviceModel"] = "room"
             };
-            var state = new Dictionary<string, object>
+            var sensors = new Dictionary<string, object>
```
dluc: please revert this, generally speaking the `state` variable contains also non-sensors data

Services/DeviceClient.cs
```diff
@@ -107,30 +106,8 @@ public class DeviceClient : IDeviceClient
...
-                }
-            }
-
-            // Write properties
-            var reportedProperties = DictionaryToTwinCollection(device.Twin.ReportedProperties);
-            await this.rateLimiting.LimitTwinWritesAsync(
-                () => this.client.UpdateReportedPropertiesAsync(reportedProperties));
-            */
+            // TODO branch for twin updates to IoT Hub located at:
+            //      https://github.com/Azure/device-simulation-dotnet/tree/send-twin-updates
```
dluc: what's the difference between this PR and that branch?

Services/Models/InternalDeviceProperties.cs
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        public object Get(string key)
+        {
+            return this.dictionary[key];
+        }
+
+        public Dictionary<string, object> GetAll()
+        {
+            return new Dictionary<string, object>(this.dictionary);
```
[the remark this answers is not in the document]
dluc: @jillcary I suppose the intent is to report a copy? if that's the case, please add a test to ensure it is a copy

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -84,16 +93,21 @@ public class JavascriptInterpreter : IJavascriptInterpreter
...
-                this.log.Debug("JS function success", () => new { filename, result });
-                return result;
+                JsValue output = engine.Execute(program).Invoke(
+                    "main",
+                    context,
+                    this.deviceState.GetAll(),
+                    this.deviceProperties.GetAll());
+
+                // update the internal device state with the new state
+                this.UpdateState(output);
```
dluc: now there are 2 ways of updating the device state, either `main()` returns a state or it calls `updateState()`. Do you think we should drop the former?

Services/Simulation/ScriptInterpreter.cs
```diff
@@ -9,10 +9,18 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Simulation
 {
     public interface IScriptInterpreter
     {
-        Dictionary<string, object> Invoke(
+        /// <summary>Invoke one of the device script files</summary>
+        /// <param name="script">Name of the script</param>
+        /// <param name="context">Context, e.g. current time, device Id, device Model</param>
+        /// <param name="state">Current device sensors state</param>
```
dluc: nit: not just sensors, the device state can hold simulation state values unrelated to sensors

Services/Simulation/ScriptInterpreter.cs
```diff
@@ -9,10 +9,18 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Simulation
...
+        /// <param name="properties">Current device properties state</param>
+        /// <remarks> Updates the internal device sensors state and internal device properties</remarks>
```
dluc: nit: remove leading space

Services/Simulation/ScriptInterpreter.cs
```diff
@@ -9,10 +9,18 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Simulation
...
+        void Invoke(
             Script script,
             Dictionary<string, object> context,
-            Dictionary<string, object> state);
+            IInternalDeviceState state,
+            IInternalDeviceProperties properites
+            );
```
dluc: nit: move `);` at the end of the previous line

SimulationAgent/SimulationRunner.cs
```diff
@@ -411,5 +450,17 @@ private void TryToStopStateThread()
...
+
+        private void TryToStopPropertiesThread()
+        {
+            try
+            {
+                this.devicesPropertiesThread.Interrupt();
+            }
+            catch (Exception e)
+            {
+                this.log.Warn("Unable to stop the devices state thread in a clean way", () => new { e });
```
dluc: fix sentence

Services/Simulation/InternalInterpreter.cs
jillcary: Yes. The Invoke method already modifies the value passed in and so doesn't also need to return the value and assign it.

Services/Simulation/JavascriptInterpreter.cs
jillcary: It feels more intuitive to me to return a state from main() rather than requiring the script to call updateState() before exiting. But certainly, from a functional perspective calling updateState() would serve the same purpose.

Services/Models/SmartDictionary.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    public class SmartDictionary : ISmartDictionary
+    {
+        private IDictionary<string, object> dictionary;
+
+        public bool Changed { get; private set; }
+
+        public SmartDictionary()
+        {
+            this.dictionary = new ConcurrentDictionary<string, object>();
```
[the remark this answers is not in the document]
dluc: ping

Services/data/devicemodels/scripts/chiller-01-state.js
```diff
@@ -19,23 +20,35 @@ var state = {
...
  *
- * @param previousState The output of main() from the previous iteration
+ * @param previousState device state from the previous iteration
+ * @param previousProperties device properties from the previous iteration
  */
-function restoreState(previousState) {
+function restoreSimulation(previousState, previousProperties) {
     // If the previous state is null, force a default state
-    if (previousState !== undefined && previousState !== null) {
+    if (previousState) {
```
dluc: did you verify whether this syntax is equivalent to `if (previousState !== undefined && previousState !== null) {`

Services/Simulation/ScriptInterpreter.cs
```diff
@@ -9,10 +9,18 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Simulation
...
+        /// <param name="state">Current device state</param>
+        /// <param name="properties">Current device properties</param>
+        /// <remarks> Updates the internal device state and internal device properties</remarks>
+        void Invoke(
             Script script,
             Dictionary<string, object> context,
-            Dictionary<string, object> state);
+            ISmartDictionary state,
+            ISmartDictionary properties
+            );
```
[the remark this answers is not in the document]
dluc: ping

SimulationAgent/DeviceState/DeviceStateActor.cs
```diff
@@ -145,27 +145,41 @@ private void MoveForward()
...
-            // put reported properties from device model into state
-            foreach (var property in deviceModel.Properties)
-                state.Add(property.Key, property.Value);
+            return properties;
+        }
+
+        /// <summary>
+        /// Initializes device state from the device model.
+        /// </summary>
+        private ISmartDictionary InitializeTelemetry(DeviceModel model)
```
dluc: nit: rename to `InitializeState`

SimulationAgent/DeviceState/DeviceStateActor.cs
```diff
@@ -145,27 +145,41 @@ private void MoveForward()
             throw new Exception("Application error, MoveForward() should not be invoked when status = " + this.status);
         }
 
-        private Dictionary<string, object> SetupTelemetryAndProperties(DeviceModel deviceModel)
+        /// <summary>
+        /// Initializes device properties from the device model.
+        /// </summary>
+        private ISmartDictionary InitializeProperties(DeviceModel model)
```
dluc: nit: rename to `GetInitialProperties`

Services/Simulation/InternalInterpreter.cs
```diff
@@ -16,12 +17,13 @@ public interface IInternalInterpreter
         /// <param name="scriptPath">Name of the script</param>
         /// <param name="scriptParams">Script parameters, e.g. min, max, step</param>
         /// <param name="context">Context, e.g. current time, device Id, device Model</param>
-        /// <param name="state">Current device sensors state</param>
-        /// <returns>New device sensors state</returns>
-        Dictionary<string, object> Invoke(
+        /// <param name="state">Current device sensors state and device properties state</param>
+        /// <remarks> Updates the internal device sensors state </remarks>
```
dluc: nit: remove the trailing spaces

Services.Test/SmartDictionaryTest.cs
```diff
@@ -0,0 +1,264 @@
...
+
+            // Act
+            var result = this.target.Has(KEY);
+
+            // Assert
+            Assert.False(result);
+        }
+
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void Changed_SHould_Return_True_When_New_Property_Added()
```
ppathan: nit: typo ```SHould```

Services/data/devicemodels/scripts/chiller-01-state.js
ppathan: Its pretty common practice in javascript to check for both undefined and null like this. https://developer.mozilla.org/en-US/docs/Glossary/Truthy

jillcary: Yes, since undefined and null are falsey values in JS this is equivalent.

Services/data/devicemodels/scripts/elevator-01-state.js
ppathan: I think you missed this one.

Services/data/devicemodels/scripts/chiller-01-state.js
dluc: There are differences between `null` and `truthy`:
```
var x = 0;
if (x !== undefined && x !== null) { use the value /* this will run */ }
if (x)                                               { use the value /* this will not run */ }
```

The previous code allows to use the state to store an empty string, a `false`, a `0`, a `null` etc.

The new code would discard those values and replace them with the default state.

PR #178, opened 2018-03-14
Docker container excessively privileged running root user

zhang-hua:
* Add default non-root user 'pcsuser'
* Run service as 'pcsuser'

PBI[2211778]

# Type of change? 

- [] Bug fix
- [ ] New feature
- [x] Enhancement

Without explicitly defining a running user in a Dockerfile definition, the root user is utilized by default.
Each respective Dockerfile should explicitly define a user and mandate that user to be the service runner via the USER instruction. Further, the service user should also have a defined GROUP. Without one, the primary group for the user will fall back to the root group.

**Checklist:**

- [x] All tests passed

### DIFF
--- a/scripts/docker/Dockerfile
+++ b/scripts/docker/Dockerfile
@@ -4,7 +4,12 @@ MAINTAINER Devis Lucato (https://github.com/dluc)
 
 LABEL Tags="Azure,IoT,Solutions,Simulation,.NET"
 
+ARG user=pcsuser
+
+RUN useradd -m -s /bin/bash -U $user
+
 COPY . /app/
+RUN chown -R $user.$user /app
 WORKDIR /app
 
 RUN \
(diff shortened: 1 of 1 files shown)


scripts/docker/Dockerfile
```diff
@@ -4,7 +4,12 @@ MAINTAINER Devis Lucato (https://github.com/dluc)
 
 LABEL Tags="Azure,IoT,Solutions,Simulation,.NET"
 
+ARG user=pcsuser
```
dluc: I think we changed this to another name, right?

ppathan: I didn't check all the repos but few and we are using pcsuser, i think it was something else before.

zhang-hua: we use pcsuser initially before using app.  Maybe we can change it to be consistent.

PR #180, opened 2018-03-22
Add DevicePropertiesActor and push reported properties updates to the IoT Hub

jillcary:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [x] Enhancement

In order to send new reported properties from the `DeviceModel` and update those properties as needed from the `DeviceStateActor` for each device, this change adds the `DevicePropertiesActor` that is run on a separate thread in order to push reported properties updates as needed. 

**NOTE: This change **does not** include receiving Desired Properties change notifications**

### This change includes:
* Adds a new actor called DevicePropertiesActor for sending updates to the IoTHub
* Creates and manages new Actor in the SimulationRunner
* DeviceProperties/UpdateReportedProperties logic for calling the device client to send the changes
* Updates the DeviceClient to include the UpdatePropertiesAsync method to update the hub reported properties
* Removes the old UpdateReportedProperties commented out code
* Tests for UpdateReportedProperties and DevicePropertiesActor
* Adds properties back into existing device models

**Checklist:**

- [x] All tests passed


Services/Models/SmartDictionary.cs
```diff
@@ -0,0 +1,101 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
```
timlaverty: I don't really  understand why we added this class.  I tlooks like a thin wrapper around ConcurentDictionary with the addition of a Changed prop.  Does it add enough value to keep it as a standalone class?

Services/Simulation/InternalInterpreter.cs
```diff
@@ -47,43 +50,43 @@ public InternalInterpreter(ILogger logger)
...
         // For each sensors specified, generate a random number in the range requested
-        private Dictionary<string, object> RunRandomNumberScript(object scriptParams, Dictionary<string, object> state)
+        private void RunRandomNumberScript(object scriptParams, ISmartDictionary state)
         {
             var sensors = this.JsonParamAsDictionary(scriptParams);
             foreach (var sensor in sensors)
             {
                 (double min, double max) = this.GetMinMaxParameters(sensor.Value);
-                state[sensor.Key] = this.random.NextDouble() * (max - min) + min;
+                var value = this.random.NextDouble() * (max - min) + min;
```
timlaverty: why var value = and not just set to the state directly like before?

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -46,25 +50,30 @@ public class JavascriptInterpreter : IJavascriptInterpreter
         /// <summary>
         /// Load a JS file and execute the main() function, passing in
         /// context information and the output from the previous execution.
-        /// Returns a map of values.
+        /// Modifies the internal device state with the latest values.
```
timlaverty: I don't think it always modifies internal device state... it can though.

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -84,16 +93,21 @@ public class JavascriptInterpreter : IJavascriptInterpreter
...
                 this.log.Debug("Executing JS function", () => new { filename });
-                JsValue output = engine.Execute(program).Invoke("main", context, this.deviceState);
 
-                var result = this.JsValueToDictionary(output);
-                this.log.Debug("JS function success", () => new { filename, result });
-                return result;
+                JsValue output = engine.Execute(program).Invoke(
+                    "main",
+                    context,
+                    this.deviceState.GetAll(),
```
timlaverty: I'm assuming state & properties are passed byval - will the script potentially update state & properties directly (other than throught he callbacks)?

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -178,26 +192,34 @@ private void UpdateState(JsValue data)
...
-            Dictionary<string, object> stateChanges;
+            Dictionary<string, object> stateChanges = this.JsValueToDictionary(data);
 
             this.log.Debug("Updating state from the script", () => new { data, this.deviceState });
 
-            stateChanges = this.JsValueToDictionary((JsValue) data);
-
-            //Update device state with the script data passed
+            // Update device state with the script data passed
             lock (this.deviceState)
```
timlaverty: do you need the lock anymore?  You're using a concurrent dictionary for this now.

timlaverty: do you need the lock anymore? You're using a concurrent dictionary for this now.

Services/Simulation/JavascriptInterpreter.cs
```diff
@@ -178,26 +192,34 @@ private void UpdateState(JsValue data)
...
-                        this.log.Debug("state change", () => new { key, value });
-                        this.deviceState[key] = value;
-                    }
+                    this.log.Debug("state change", () => new { key, value });
+                    this.deviceState.Set(key, value);
                 }
             }
         }
+
+        // TODO: Move this out of the scriptinterpreter class into DeviceStateActor to keep this class stateless
```
timlaverty: Can you move this out of the class with this PR?

timlaverty:
Let's review F2F - the main thing I'm missing is understanding how you're hooking up property change events from hub -> simulated devices.  It could be in a different PR which is fine.

This PR looks like it's managing property changes made from method execution in the simulated device.

Services/Simulation/JavascriptInterpreter.cs
jillcary: That's the intention behind the wrapper for the concurrent dictionary, which should only allow for updates to the actual state via a state.Set()

Services/Simulation/JavascriptInterpreter.cs
jillcary: Good catch, will remove.

Services/Simulation/JavascriptInterpreter.cs
jillcary: I opted not to do it in this PR to limit the scope for the original RMv2 functionality adds.

Services/Models/SmartDictionary.cs
jillcary: As you noted below, the state and properties are passed by value to the script interpreter. This class, on a get should return a new dictionary. The way the scripts modify the internal state would be through the Set method.

This could be renamed to DeviceProperties and then we just use a ConcurrentDictionary for state, but since they are so similar it made the most sense for the method calls to have a similar pattern.

Services/Simulation/InternalInterpreter.cs
jillcary: It was intended to match the style of the other methods below.

Services/Simulation/JavascriptInterpreter.cs
jillcary: I believe the call to `this.UpdateState(output);` changes this behavior. The return values is now void.

Services/Models/SmartDictionary.cs
jillcary: Also important to note that this was reviewed and discussed in https://github.com/Azure/device-simulation-dotnet/pull/173

Due to waiting for the release these look like net new changes

SimulationAgent/DeviceState/DeviceStateActor.cs
```diff
@@ -25,13 +27,10 @@ public enum ActorStatus
             Updating
         }
 
-        public const string CALC_TELEMETRY = "CalculateRandomizedTelemetry";
+        public ISmartDictionary DeviceState { get; set; }
```
ppathan: Should we include the summary?
`     /// <summary> `
`        /// The virtual state of the simulated device. The state is`
`        /// periodically updated using an external script.`
`        /// </summary>`

SimulationAgent.Test/DeviceState/DeviceStateActorTest.cs
```diff
@@ -66,18 +64,7 @@ private void SetupDeviceStateActor()
...
-            };
-
-            this.scriptInterpreter
-                .Setup(x => x.Invoke(
-                    It.IsAny<Script>(),
-                    It.IsAny<Dictionary<string, object>>(),
-                    It.IsAny<Dictionary<string, object>>()))
-                .Returns(deviceState);
-
+
```
ppathan: Do we not need the mock setup for this test?

Services/Diagnostics/ActorsLogger.cs
```diff
@@ -77,7 +77,7 @@ public void Setup(string deviceId, string actorName)
             this.telemetryLogFile = this.path + Path.DirectorySeparatorChar + "telemetry.log";
 
             this.enabled = this.enabledInConfig && !string.IsNullOrEmpty(this.path);
-            
+
```
ppathan: nit: Why newline?

Services.Test/SmartDictionaryTest.cs
```diff
@@ -0,0 +1,264 @@
...
+        private ISmartDictionary target;
+
+        public SmartDictionaryTest(ITestOutputHelper log)
+        {
+            // Initialize device properties
+            this.target = this.GetTestProperties();
+        }
+
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void Should_Be_Empty_On_Get_All_When_No_Properties_Added()
```
ppathan: The naming convention can be improved. So when following `Should_ExpectedBehavior_When_StateUnderTest`. It makes sense to have `Should_BeEmpty_When_GetAllEmpty`

Services.Test/SmartDictionaryTest.cs
```diff
@@ -0,0 +1,264 @@
...
+
+            // Act
+            var props = this.target.GetAll();
+
+            // Assert
+            Assert.Empty(props);
+        }
+
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void Should_Return_All_Test_Properties_On_Get_All_When_Initialized_With_Device_Model()
```
ppathan: Following same convention: `Should_ReturnAllTestProperties_When_InitializedWithDeviceModel`

Services.Test/SmartDictionaryTest.cs
```diff
@@ -0,0 +1,264 @@
...
+            Assert.Equal("TestChiller", props["Type"]);
+            Assert.Equal("1.0", props["Firmware"]);
+            Assert.Equal("TestCH101", props["Model"]);
+            Assert.Equal("TestBuilding 2", props["Location"]);
+            Assert.Equal(47.640792, props["Latitude"]);
+            Assert.Equal(-122.126258, props["Longitude"]);
+        }
+
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void Should_Throw_On_Get_When_Key_Does_Not_Exist()
```
ppathan: `Should_ThrowException_When_KeyDoesNotExist`

jillcary:
Udpate: Abandoning this PR as it has the wrong base branch and this is a duplicate of https://github.com/Azure/device-simulation-dotnet/pull/173 

Please see https://github.com/Azure/device-simulation-dotnet/pull/181 for the correct branch comparison.

PR #184, opened 2018-03-28
Register for Desired Property State Changes

jillcary:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [x] Enhancement

Adds functionality for receiving notifications of Desired Property state changes from the IoT Hub. When the hub detects a state change it will send a notification via the `DeviceClient.SetDesiredPropertyUpdateCallbackAsync` registration. This change allows the simulation service to subscribe to the change notifications and update the internal simulation state to report the desired property to the hub when executed.

*Note:*
This registration is different from the way desired properties changes happened in RemoteMontioring V2 Preview, which was previously querying for changes on a set interval.

#### This changes includes:

| File       | Description           |
| ------------- | ------------- |
| DeviceClient.cs |Add registration for desired properties callback |
| _[New]_ DevicePropertiesRequest.cs | Calls the azure device client to set desired properties update callback and includes the logic to execute on callback to update the internal device state |
| _[New]_ DevicePropertiesRequestTest.cs | Unit tests for `DevicePropertiesRequest`      |
| Connect.cs | Add registration for desired properties change events during the connect logic in the `DeviceConnectionActor` |

**Checklist:**

- [x] All tests passed

### DIFF
--- a/Services.Test/DevicePropertiesRequestTest.cs
+++ b/Services.Test/DevicePropertiesRequestTest.cs
@@ -40,7 +40,7 @@ namespace Services.Test
         }
 
         [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
-        public void SmartDictionary_Should_Update_When_DesiredPropertiesChange()
+        public void SmartDictionary_Should_UpdateValue_When_DesiredPropertiesChange()
         {
             // Arrange
             const string NEW_VALUE = "new value";
@@ -59,6 +59,49 @@ namespace Services.Test
             Assert.Equal(result, NEW_VALUE);
         }
 
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void SmartDictionary_Should_HaveNewItem_When_NewDesiredPropertyAdded()
+        {
+            // Arrange
+            const string NEW_KEY = "new key";
+            const string NEW_VALUE = "new value";
+
+            ISmartDictionary reportedProps = GetTestProperties();
+            this.target.RegisterDevicePropertiesUpdateAsync(DEVICE_ID, reportedProps);
+
+            TwinCollection desiredProps = new TwinCollection();
+            desiredProps[NEW_KEY] = NEW_VALUE;
+
+            // Act
+            this.target.OnPropertyUpdateRequested(desiredProps, null);
+            var result = reportedProps.Get(NEW_KEY);
+
+            // Assert
+            Assert.Equal(result, NEW_VALUE);
+        }
+
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void SmartDictionary_Should_Not_Update_When_DesiredPropertiesValueIsTheSame()
+        {
+            // Arrange
+            ISmartDictionary reportedProps = GetTestProperties();
+            reportedProps.ResetChanged();
+            Assert.False(reportedProps.Changed);
+
+            this.target.RegisterDevicePropertiesUpdateAsync(DEVICE_ID, reportedProps);
+
+            TwinCollection desiredProps = new TwinCollection
+            {
+                [KEY1] = VALUE1 // This should be the same value in props
+            };
+
+            // Act
+            this.target.OnPropertyUpdateRequested(desiredProps, null);
+            var result = reportedProps.Get(KEY1);
+
+            // Assert
+            Assert.False(reportedProps.Changed);
+        }
 
         private SdkClient GetSdkClient()
         {
--- a/Services.Test/DevicePropertiesRequestTest.cs
+++ b/Services.Test/DevicePropertiesRequestTest.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Shared;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models;
+using Moq;
+using Services.Test.helpers;
+using Xunit;
+using Xunit.Abstractions;
+using SdkClient = Microsoft.Azure.Devices.Client.DeviceClient;
+
+namespace Services.Test
+{
+    public class DevicePropertiesRequestTest
+    {
+        private const string DEVICE_ID = "01";
+        private const string KEY1 = "Key1";
+        private const string KEY2 = "Key2";
+        private const string VALUE1 = "Value1";
+        private const string VALUE2 = "Value2";
+
+        private Mock<IDeviceClient> client;
+        private SdkClient sdkClient;
+        private Mock<ILogger> logger;
+
+        private IDevicePropertiesRequest target;
+
+        public DevicePropertiesRequestTest(ITestOutputHelper log)
+        {
+            this.sdkClient = GetSdkClient();
+
+            this.client = new Mock<IDeviceClient>();
+            this.logger = new Mock<ILogger>();
+
+            this.target = new DevicePropertiesRequest(sdkClient, this.logger.Object);
+        }
+
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void SmartDictionary_Should_Update_When_DesiredPropertiesChange()
+        {
+            // Arrange
+            const string NEW_VALUE = "new value";
+
+            ISmartDictionary reportedProps = GetTestProperties();
+            this.target.RegisterDevicePropertiesUpdateAsync(DEVICE_ID, reportedProps);
+
+            TwinCollection desiredProps = new TwinCollection();
+            desiredProps[KEY1] = NEW_VALUE;
+
+            // Act
+            this.target.OnPropertyUpdateRequested(desiredProps, null);
+            var result = reportedProps.Get(KEY1);
+
+            // Assert
+            Assert.Equal(result, NEW_VALUE);
+        }
+
+
+        private SdkClient GetSdkClient()
+        {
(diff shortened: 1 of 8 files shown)


Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
+{
+    public interface IDevicePropertiesRequest
+    {
+        Task RegisterDevicePropertiesUpdateAsync(
+            string deviceId,
+            ISmartDictionary deviceProperties);
+
+        Task OnPropertyUpdateRequested(
```
ppathan: OnPropertyUpdateRequestedAsync or OnPropertyUpdateRequestedCallback if not really async?

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            if (desiredProperties.Count > 0)
+            {
+                // This is where custom code for handling specific desired property changes could be added.
+                // For the purposes of the simulation service, we have chosen to write the desired properties
+                // directly to the reported properties. 
+
+                try
+                {
+
```
ppathan: nit: extra line

Services.Test/DevicePropertiesRequestTest.cs
```diff
@@ -0,0 +1,126 @@
...
+            this.target.RegisterDevicePropertiesUpdateAsync(DEVICE_ID, reportedProps);
+
+            TwinCollection desiredProps = new TwinCollection
+            {
+                [KEY1] = VALUE1 // This should be the same value in props
+            };
+
+            // Act
+            this.target.OnPropertyUpdateRequested(desiredProps, null);
+            var result = reportedProps.Get(KEY1);
```
ppathan: nit : result is not used

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,98 @@
...
+            this.log.Debug("Callback for desired properties updates setup successfully", () => new { this.deviceId });
+        }
+
+        /// <summary>
+        /// When a desired property change is requested, update the internal the device state properties
+        /// which will be reported to the hub. If there is a new desired property that does not exist in
+        /// the reported properties, it will be added.
+        /// </summary>
+        public Task OnPropertyUpdateRequested(TwinCollection desiredProperties, object userContext)
```
avanikp: When/where will userContext be used ?

jillcary: It is not used, but is the required format for the SetDesiredPropertyUpdateCallbackAsync method. ([documentation link](https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/iot-edge/tutorial-csharp-module.md))

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            this.log.Info("Desired property update requested", () => new { this.deviceId, desiredProperties });
+
+            if (desiredProperties.Count > 0)
```
avanikp: can desiredProperties be null ? Do we need a check for null ?

SimulationAgent/DeviceConnection/Connect.cs
```diff
@@ -49,6 +49,8 @@ public void Setup(IDeviceConnectionActor context, string deviceId, DeviceModel d
                 var now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
                 await this.context.Client.ConnectAsync();
                 await this.context.Client.RegisterMethodsForDeviceAsync(this.deviceModel.CloudToDeviceMethods, this.context.DeviceState, this.context.DeviceProperties);
+                await this.context.Client.RegisterDesiredPropertiesUpdateAsync(this.context.DeviceProperties);
```
avanikp: nit: should we consider 'Updates' (plural) instead of 'Update' in the name ? since this is per device should it be ResisterDesiredPropertiesUpdatesForDeviceAsync or something similar ?

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,96 @@
...
+            this.deviceId = string.Empty;
+        }
+
+        public async Task RegisterDevicePropertiesUpdateAsync(string deviceId, ISmartDictionary deviceProperties)
+        {
+            if (this.deviceId != string.Empty)
+            {
+                this.log.Error("Application error, each device must have a separate instance", () => { });
+                throw new Exception("Application error, each device must have a separate instance of " + this.GetType().FullName);
+            }
```
avanikp: curious, how would the setup get into this state ? Is this check necessary ?

SimulationAgent/DeviceConnection/Connect.cs
jillcary: Interesting to consider. I was actually wondering if we should keep "property" singular based on the DeviceClient API: `SetDesiredPropertyUpdateCallbackAsync`

Since this isn't the callback though, it does make sense to me to make it plural for "updates"

Services/DevicePropertiesRequest.cs
jillcary: It's a preventative measure in case the consumer of this class were to call registration more than once. Each device should have a separate instance of `DeviceClient` which has `DevicePropertiesRequest` that does the registration with the azure device client only once. If the deviceId were already set then this class could accidentally update the state on the callback for the wrong device on the hub.

SimulationAgent/DeviceConnection/Connect.cs
ppathan: My 2 cents.
I like what you have since these are going to be called per device any ways, so it is update per device. But i'm ok with whatever you decide.

SimulationAgent/DeviceState/DeviceStateActor.cs
```diff
@@ -152,6 +153,10 @@ private ISmartDictionary GetInitialProperties(DeviceModel model)
         {
             var properties = new SmartDictionary();
 
+            // Add SupportedMethods property with methods listed in device model
+            properties.Set(SUPPORTED_METHODS_KEY, string.Join(",", this.deviceModel.CloudToDeviceMethods.Keys));
```
ppathan: Yay!

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <summary>
+        /// When a desired property change is requested, update the internal the device state properties
+        /// which will be reported to the hub. If there is a new desired property that does not exist in
+        /// the reported properties, it will be added.
+        /// </summary>
+        public Task OnPropertyUpdateRequestedCallback(TwinCollection desiredProperties, object userContext)
+        {
+            this.log.Info("Desired property update requested", () => new { this.deviceId, desiredProperties });
+
+            if (desiredProperties != null && desiredProperties.Count > 0)
```
timlaverty: Why do you need to do this?  If it fires aren't you guaranteed to have changes?  If it's firing for tag updates or reported property updates (I think I remember seeing this) can you file a big on the platform and link it here?

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                // This is where custom code for handling specific desired property changes could be added.
+                // For the purposes of the simulation service, we have chosen to write the desired properties
+                // directly to the reported properties. 
+                try
+                {
+                    foreach (KeyValuePair<string, object> item in desiredProperties)
+                    {
+                        // Only update if key doesn't exist or value has changed 
+                        if (!this.deviceProperties.Has(item.Key) ||
```
timlaverty: is deviceProperties the container for reported properties?

timlaverty: What happens if a new desired property is added?  this looks like it only updates existing values & doesn't add new properties, correct?

SimulationAgent/DeviceState/DeviceStateActor.cs
timlaverty: Who is using this?  Is it for the UI to enumerate which methods are supported for devices?

Services/DevicePropertiesRequest.cs
jillcary: Yes, passed into the constructor as the `ISmartDictionary` from the `DeviceStateActor`.

jillcary: `!this.deviceProperties.Has(item.Key)` will evaluate as true and `.Set()` will be called which has the behavior of updating existing properties OR creating new properties if it doesn't exist. (see comment below)

Also see unit test `SmartDictionary_Should_HaveNewItem_When_NewDesiredPropertyAdded()` in `DevicePropertiesRequestTest.cs`

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,91 @@
...
+            this.log.Debug("Setting up callback for desired properties updates.", () => new { this.deviceId });
+
+            // Set callback that IoT Hub calls whenever the client receives a desired properties state update.
+            await this.client.SetDesiredPropertyUpdateCallbackAsync(OnPropertyUpdateRequestedCallback, null);
+
+            this.log.Debug("Callback for desired properties updates setup successfully", () => new { this.deviceId });
+        }
+
+        /// <summary>
+        /// When a desired property change is requested, update the internal the device state properties
```
timlaverty: nit: "update the internal the device"

SimulationAgent/DeviceState/DeviceStateActor.cs
timlaverty: This seems like it should expose everything you need to describe to the client the function: name of function, parameters, return value, etc.  The UI will need to show a combo box for function name but then it's going to need to build UI for parameters at least isn't it?

timlaverty: I'd picture  JSON schema that we put in the twin that offers what you need - suggest talking with @dluc about device models.

jillcary: This is copied directly from: https://github.com/Azure/device-simulation-dotnet/blob/0d37f326717db065dab767105763d29a0ece24bd/SimulationAgent/Simulation/DeviceStatusLogic/DeviceBootstrap.cs#L140



Added a new issue to track the new feature request:
https://github.com/Azure/device-simulation-dotnet/issues/191

PR #182, opened 2018-03-25
Parvezp/add methods

ppathan:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Breaking change (breaks backward compatibility)

# Description, Context, Motivation 
Enabling methods for the new simulation brancx

**Checklist:**

- [x] All tests passed
- [x] The code follows the code style and conventions of this project
- [ ] The change requires a change to the documentation
- [ ] I have updated the documentation accordingly

PR #193, opened 2018-04-03
Add Device Properties Functionality

jillcary:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

This is the merge of the following PRs and branches for re-adding device properties support back into the latest simulation branch. 

- [Add Device Properties logic to internal simulation](https://github.com/Azure/device-simulation-dotnet/pull/173)
- [Add DevicePropertiesActor and push reported properties updates to the IoT Hub](https://github.com/Azure/device-simulation-dotnet/pull/181)
- [Register for Desired Property State Changes](https://github.com/Azure/device-simulation-dotnet/pull/184)
- [Firmware update method calls updateProperty callback with all parameters](https://github.com/Azure/device-simulation-dotnet/pull/186)
- [Add Methods](https://github.com/Azure/device-simulation-dotnet/pull/182) 
- [Parvezp/add methods](https://github.com/Azure/device-simulation-dotnet/pull/182)

**Checklist:**

- [x] All tests passed


Services/Models/SmartDictionary.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+            else
+            {
+                this.dictionary.Add(key, value);
+            }
+
+            this.Changed = true;
+        }
+
+        public void SetAll(Dictionary<string, object> newState)
```
dluc: I'd force this to be `ConcurrentDictionary`

jillcary: Done.

SimulationAgent/DeviceProperties/DevicePropertiesActor.cs
```diff
@@ -0,0 +1,211 @@
...
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent.Exceptions;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent.DeviceProperties
+{
+    public interface IDevicePropertiesActor
+    {
+        ISmartDictionary DeviceProperties { get; }
+        ISmartDictionary DeviceState { get; }
+        IDeviceClient Client { get; }
+        Device Device { get; }
```
dluc: looks like this is not used, can we remove it?

SimulationAgent/DeviceProperties/DevicePropertiesActor.cs
```diff
@@ -0,0 +1,211 @@
...
+
+        /// <summary>
+        /// Azure IoT Hub client
+        /// </summary>
+        public IDeviceClient Client => this.deviceConnectionActor.Client;
+
+        /// <summary>
+        /// Azure IoT Hub Device instance
+        /// </summary>
+        public Device Device => this.deviceConnectionActor.Device;
```
dluc: looks like this is not used, can we remove it?

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,91 @@
...
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
+{
+    public interface IDevicePropertiesRequest
+    {
+        Task RegisterDevicePropertyUpdatesAsync(
+            string deviceId,
+            ISmartDictionary deviceProperties);
+
+        Task OnPropertyUpdateRequestedCallback(
```
dluc: this method is not used outside the class below, can we then remove it from the interface and make it private?

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,91 @@
...
+            this.log.Debug("Setting up callback for desired properties updates.", () => new { this.deviceId });
+
+            // Set callback that IoT Hub calls whenever the client receives a desired properties state update.
+            await this.client.SetDesiredPropertyUpdateCallbackAsync(OnPropertyUpdateRequestedCallback, null);
+
+            this.log.Debug("Callback for desired properties updates setup successfully", () => new { this.deviceId });
+        }
+
+        /// <summary>
+        /// When a desired property change is requested, update the internal the device state properties
```
dluc: typo

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,91 @@
...
+                    // Only update if key doesn't exist or value has changed 
+                    if (!this.deviceProperties.Has(item.Key) ||
+                        (item.Value.ToString() != this.deviceProperties.Get(item.Key).ToString()))
+                    {
+                        // Update existing property or create new property if key doesn't exist.
+                        this.deviceProperties.Set(item.Key, item.Value);
+                    }
+                }
+            }
+            catch (Exception e)
```
dluc: when could this happen?

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                this.log.Error("Error updating internal device state properties", () => new { e, this.deviceId, desiredProperties });
+            }
+
+            this.log.Debug("Desired property update successfully reported to internal state", () => new { this.deviceId, desiredProperties });
```
dluc: could you rephrase? something like "device properties updated to the desired values"

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,91 @@
...
+        public DevicePropertiesRequest(Azure.Devices.Client.DeviceClient client, ILogger logger)
+        {
+            this.client = client;
+            this.log = logger;
+            this.deviceId = string.Empty;
+        }
+
+        public async Task RegisterDevicePropertyUpdatesAsync(string deviceId, ISmartDictionary deviceProperties)
+        {
+            if (this.deviceId != string.Empty)
```
dluc: I would pass the deviceId to the ctor class, and have a "registered" boolean for this purpose

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,91 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Shared;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
+{
+    public interface IDevicePropertiesRequest
```
dluc: the name is a bit strange, e.g. why "request" ? I would have chosen something like 
```
IDeviceProperties {
    Task OnChangeUpdateAsync(ISmartDictionary deviceProperties)
}
```

Services/DevicePropertiesRequest.cs
```diff
@@ -0,0 +1,91 @@
...
+            // This is where custom code for handling specific desired property changes could be added.
+            // For the purposes of the simulation service, we have chosen to write the desired properties
+            // directly to the reported properties. 
+            try
+            {
+                foreach (KeyValuePair<string, object> item in desiredProperties)
+                {
+                    // Only update if key doesn't exist or value has changed 
+                    if (!this.deviceProperties.Has(item.Key) ||
+                        (item.Value.ToString() != this.deviceProperties.Get(item.Key).ToString()))
```
dluc: wondering if `ToString` is the right way of comparing values. For instance, what if the value is a complex object? Does `ToString` serialize the actual value?

jillcary: I used it because it is the comparison type used in the previous implementation that was proven to work:
https://github.com/Azure/device-simulation-dotnet/blob/0d37f326717db065dab767105763d29a0ece24bd/SimulationAgent/Simulation/DeviceStatusLogic/UpdateReportedProperties.cs#L160

However, ToString is overall not very efficient for comparison because every property's value is accessed. I thought of possibly implementing a comparer but because the objects are generic, couldn't think of a better way to go about it. 

Any alternate suggestions?

Services/DevicePropertiesRequest.cs
jillcary: The implementation calls dictionary.Add which can throw an ArgumentNull exception

Services/DevicePropertiesRequest.cs
dluc: I think that "every value is accessed" is a requirement in order to compare the value, right?

In this before-vs-after example, to detect that "five" was added under the "someproperty" property,
the code needs to see all the props:

```
"desired-properties": {
	"someproperty": {
	    "one": {
	       "two": 2,
	       "three": 3
	    },
	    "four": [ 0,1,2,3 ]
	}
}
```

```
"desired-properties": {
	"someproperty": {
	    "one": {
	       "two": 2,
	       "three": 3,
	       "five": 5
	    },
	    "four": [ 0,1,2,3 ]
	}
}
```

jillcary: Since this meets the parity requirements for RMv2 Preview, I've filed a bug here that we can address:
https://github.com/Azure/device-simulation-dotnet/issues/194

PR #195, opened 2018-04-05
Add IsSimulated Tag when devices are created from list

jillcary:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

From the looks of it, in the `DeviceConnectionActor` the state goes from:
**Started>FetchCompleted>Connected**

This means it never enters the **Started>FetchCompleted>DeviceRegistered > DeviceTwinTagged>Connected** flow.

Because of the call to Devices.CreateListAsync, the registration state is never hit.
```
    // This will ignore existing devices, creating only the missing ones
    this.devices.CreateListAsync(devices)
    .Wait(TimeSpan.FromSeconds(DEVICES_CREATION_TIMEOUT_SECS));
```

This change moved is IsSimulated tag logic from the Connect Logic to the Properties Actor. The properties actor will tag the device before it starts updating the twin Reported Properties. 

**Checklist:**

- [x] All tests passed


SimulationAgent/DeviceConnection/DeviceConnectionActor.cs
```diff
@@ -124,7 +116,11 @@ public enum ActorEvents
         /// <summary>
         /// Failed device twin updates counter
         /// </summary>
-        public long FailedTwinUpdatesCount => this.failedTwinUpdatesCount;
+        public long FailedTwinUpdatesCount
+        {
+            get => this.failedTwinUpdatesCount;
+            set => this.failedTwinUpdatesCount = value;
```
maple10001: @jillcary Could you help me understand why a setter needed for `FailedTwinUpdatesCount` ?

SimulationAgent/DeviceProperties/Tag.cs
```diff
@@ -1,52 +1,50 @@
...
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models;
 
-namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent.DeviceConnection
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent.DeviceProperties
 {
     /// <summary>
     /// Add twin information to the new device twin
     /// </summary>
-    public class DeviceTwinTag : IDeviceConnectionLogic
+    public class Tag : IDevicePropertiesLogic
```
dluc: could use a longer name, this is a bit ambiguous, e.g. `TagDevice` or `SetDeviceTag` etc. By the way, if you don't want to use the term "twin" I assume you should also rename the event `DeviceTwinTagged` right ?

SimulationAgent/DeviceProperties/Tag.cs
```diff
@@ -1,52 +1,50 @@
...
 
-        public async Task RunAsync()
+        public void Run()
         {
             this.log.Debug("Adding tag to device twin...", () => new { this.deviceId });
             try
             {
                 var now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
-                await this.devices.AddTagAsync(this.deviceId);
+                this.devices.AddTagAsync(this.deviceId);
```
dluc: HandleEvent should be called when this is complete, see `SendTelemetry.cs` use of `ContinueWith`

SimulationAgent/DeviceProperties/Tag.cs
```diff
@@ -1,52 +1,50 @@
...
 
-        public void Setup(IDeviceConnectionActor context, string deviceId, DeviceModel deviceModel)
+        public void Setup(IDevicePropertiesActor context, string deviceId)
         {
             this.context = context;
             this.deviceId = deviceId;
         }
 
-        public async Task RunAsync()
+        public void Run()
```
dluc: thanks for fixing this

SimulationAgent/SimulationRunner.cs
```diff
@@ -41,6 +41,9 @@ public class SimulationRunner : ISimulationRunner
         // Set of counter to optimize scheduling
         private readonly ConnectionLoopSettings connectionLoopSettings;
 
+        // Set of counters to optimize scheduling
```
dluc: the comment (and the one on line 41) seems incorrect, these are not counters, are they?

Services/Concurrency/LoopSettings.cs
```diff
@@ -35,6 +34,25 @@ public void NewLoop()
...
+
+        public PropertiesLoopSettings(IRateLimitingConfig ratingConfig)
+        {
+            this.ratingConfig = ratingConfig;
+            this.NewLoop();
+        }
+
+        public void NewLoop()
+        {
             this.SchedulableTaggings = Math.Max(1, this.ratingConfig.RegistryOperationsPerMinute / 25);
```
dluc: this seems incorrect: tagging a device is not a registry operation, so the formula should use `TwinWritesPerSecond` instead.

dluc: please add test to ensure the formula is the intended one

SimulationAgent.Test/DeviceProperties/DevicePropertiesActorTest.cs
```diff
@@ -72,22 +88,70 @@ public void Handle_Event_Should_Throw_When_Out_Of_Range()
                 () => this.target.HandleEvent(OUT_OF_RANGE_EVENT));
         }
 
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void Should_Return_CountOfFailedTwinUpdates_When_TwinUpdateFails()
```
dluc: Allow me to say that this naming convention is really weird. Typically one would choose between PascalCasing or snake_casing, not use both...

also why doesn't "CountOfFailedTwinUpdates" have the "_" ?

my suggestion would be to stick to PascalCasing, which is how we address readability everywhere

SimulationAgent/DeviceProperties/Tag.cs
```diff
@@ -1,52 +1,50 @@
...
+    public class Tag : IDevicePropertiesLogic
     {
         private readonly IDevices devices;
         private readonly ILogger log;
         private string deviceId;
-        private IDeviceConnectionActor context;
+        private IDevicePropertiesActor context;
 
-        public DeviceTwinTag(IDevices devices, ILogger logger)
+        public Tag(IDevices devices, ILogger logger)
```
dluc: no tests for this class?

SimulationAgent.Test/DeviceProperties/DevicePropertiesActorTest.cs
```diff
@@ -72,22 +88,70 @@ public void Handle_Event_Should_Throw_When_Out_Of_Range()
...
+        {
+            // Arrange
+            const int FAILED_DEVICE_TWIN_UPDATES_COUNT = 5;
+
+            //this.SetupDevicePropertiesActor();
```
dluc: remove?

SimulationAgent.Test/DeviceProperties/DevicePropertiesActorTest.cs
```diff
@@ -72,22 +88,70 @@ public void Handle_Event_Should_Throw_When_Out_Of_Range()
...
+            this.target.Setup(DEVICE_ID,
+                this.deviceStateActor.Object,
+                SetupDeviceConnectionActor(),
```
dluc: why does this test need a new connection actor, instead of using the one created in the ctor? there is a lot of test setup without comments/instructions, which will make the tests hard to maintain

SimulationAgent/DeviceConnection/DeviceConnectionActor.cs
jillcary: Instead of adding a circular dependency on the `DeviceConnectionActor` from the `DevicePropertiesActor` if we add a setter for the `FailedTwinUpdatesCount`, then the `DevicePropertiesActor` can increment the `FailedTwinUpdatesCount` in the connection actor and the connection actor can still maintain the summed error count.

Another option instead of a setter would be a public method like "IncrementFailedTwinUpdateCount".

SimulationAgent/DeviceProperties/DevicePropertiesActor.cs
```diff
@@ -81,16 +94,26 @@ public enum ActorEvents
         /// </summary>
         public IDeviceClient Client => this.deviceConnectionActor.Client;
 
+        /// <summary>
+        /// Failed device twin updates counter
+        /// </summary>
+        public long FailedTwinUpdatesCount {
+            get => this.deviceConnectionActor.FailedTwinUpdatesCount;
+            set => this.deviceConnectionActor.FailedTwinUpdatesCount = value;
```
dluc: I think it would be easier to keep the value here, and update the simulation runner (see `SimulationErrorsCount`) to pull this value directly from this actor.  So we avoid this actor writing a value into the other actor, which makes the flow hard to follow

SimulationAgent/DeviceConnection/DeviceConnectionActor.cs
dluc: @jillcary see my other comment about `SimulationErrorsCount`, I think we can avoid this setter, updating the SimulationRunner to pull error counters out of each actor

SimulationAgent.Test/DeviceProperties/DevicePropertiesActorTest.cs
ppathan: That is because of me. I suggested we use Should_ExpectedBehavior_When_StateUnderTest naming scheme. I think it is easier to read and understand what exactly we are testing.

Services/Concurrency/LoopSettings.cs
jillcary: Sure, I was just using the existing logic. To make sure the new value is correct, can you explain the logic behind the division by 25 and 10? Is that the max allowable by the hub?

SimulationAgent/SimulationRunner.cs
jillcary: Good question, just following the convention of the other loop settings comments. I'll follow up in-person to better understand what should be happening and if we should change it for all of the loop settings members.

Services/Concurrency/LoopSettings.cs
dluc: `TwinWritesPerSecond` is the max number of operations allowed before the hub starts throttling. The division is slow down the the actor further, to leave room to other threads competing for the same resource. For instance, if 100 is the quota, and if the actor is trying to tag 100 devices per second, a simulated device X trying to update the twin would have to wait for all the devices to be tagged before the device twin X is written. I would divide by 2, to leave 50% of the quota to potential simulated devices trying to update their twin.

Services.Test/Concurrency/LoopSettingsTest.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            // Arrange
+            this.SetupRateLimitingConfig();
+
+            // Act
+            this.propertiesTarget.NewLoop();
+
+            // Assert
+            // In order for other threads to be able to schedule twin opertations,
+            // value should at least 1 but less than the limit per second.
```
ppathan: nit: value should be at least

SimulationAgent.Test/DeviceProperties/DevicePropertiesActorTest.cs
```diff
@@ -72,22 +81,89 @@ public void Handle_Event_Should_Throw_When_Out_Of_Range()
...
+        public void Should_Return_CountOfFailedTwinUpdates_When_TwinUpdateFails()
+        {
+            // Arrange
+            const int FAILED_DEVICE_TWIN_UPDATES_COUNT = 3;
+            this.CreateNewDevicePropertiesActor();
+            this.SetupDevicePropertiesActor();
+            this.SetupRateLimitingConfig();
+            this.loopSettings.Object.NewLoop(); // resets SchedulableTaggings
+
+            // The constructor should initialize count as zero.
```
ppathan: nit: count to zero

SimulationAgent/DeviceProperties/SetDeviceTag.cs
```diff
@@ -1,52 +1,50 @@
...
 
-        public async Task RunAsync()
+        public void Run()
         {
             this.log.Debug("Adding tag to device twin...", () => new { this.deviceId });
             try
             {
                 var now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
-                await this.devices.AddTagAsync(this.deviceId);
+                this.devices.AddTagAsync(this.deviceId);
```
ppathan: Previously we were waiting for the async operation to finish now we are not. Any reason?

jillcary: Because the DevicePropertiesActor is on its own thread, the state machine can execute synchronously without holding up the other actor operations. The Run() operations in the other actors follow this pattern and so I also updated this class to follow that pattern here:

image


You can see another example of it in the DeviceTelemetryActor here:
https://github.com/Azure/device-simulation-dotnet/blob/7449fde2a82f74a6ec298de231c848885aaf5f05/SimulationAgent/DeviceTelemetry/SendTelemetry.cs#L39

Services/Concurrency/LoopSettings.cs
```diff
@@ -35,7 +34,31 @@ public void NewLoop()
...
+
+        public PropertiesLoopSettings(IRateLimitingConfig ratingConfig)
+        {
+            this.ratingConfig = ratingConfig;
+            this.NewLoop();
+        }
+
+        public void NewLoop()
+        {
+            // In order for other threads to be able to schedule twin opertations,
```
dluc: typo

SimulationAgent/DeviceProperties/DevicePropertiesActor.cs
```diff
@@ -16,23 +16,33 @@ public interface IDevicePropertiesActor
...
+            PropertiesLoopSettings loopSettings);
 
         string Run();
         void HandleEvent(DevicePropertiesActor.ActorEvents e);
         void Stop();
     }
 
+    /// <summary>
+    /// The Device Properties Actor is responsible for sending updates
+    /// to the IoT Hub Device Twin. This includes adding an intial tag to
```
dluc: typo

SimulationAgent/DeviceProperties/SetDeviceTag.cs
dluc: I had a comment in the previous review, suggesting to use `ContinueWith`, like in `SendTelemetry.cs`, otherwise the app could call HandleEvent (line 42) before catching the exception (I'm not sure the exceptions would be caught either)

SimulationAgent/DeviceProperties/SetDeviceTag.cs
```diff
@@ -31,21 +33,52 @@ public void Setup(IDevicePropertiesActor context, string deviceId)
...
+                this.devices
+                    .AddTagAsync(this.deviceId)
+                    .ContinueWith(t =>
+                    {
+                        var timeSpent = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - now;
 
-                this.context.HandleEvent(DevicePropertiesActor.ActorEvents.DeviceTagged);
+                        if (t.IsCanceled)
+                        {
+                            this.log.Warn("The set device tag task has been cancelled", () => new { this.deviceId, t.Exception });
```
ppathan: nit: Log messages can follow similar pattern
`Device tag task cancelled`
`Device tag task faulted`
`Device tag task completed`

PR #201, opened 2018-04-18
Implement js device methods

peterfelts:
# Description and Motivation 

Implementing device simulation javascript methods (#44):

- SetFuelLevel
- StartElevator
- StopElevator
- StartMovingDevice
- StopMovingDevice
- SetTemperature
- FillTank
- EmptyTank
- IncreaseTemperature
- DecreaseTemperature

Removed RestartEngine.
This change involved replacing some existing TODO methods, such as FillTank and EmptyTank (which were replaced with SetFuelLevel).

# Change type 

- [ ] Bug fix
- [x] New feature
- [x] Enhancement
- [ ] Breaking change (breaks backward compatibility)

**Checklist:**

- [x] All tests passed (manually tested)
- [x] The code follows the code style and conventions of this project
- [ ] The change requires a change to the documentation
- [ ] I have updated the documentation accordingly

PR #205, opened 2018-04-27
Removing the 'Restart' method from the engine device models to align …

peterfelts:
…with the spec. Also made some minor clean up to the Empty and Fill tank methods.

When I implemented the TODO device methods (#201) the spec called for removing the 'Restart' method. I removed the method file but forgot to remove a reference to it in the Engine device models.

# Change type 

- [x ] Bug fix
- [ ] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

### DIFF
--- a/Services/data/devicemodels/engine-01.json
+++ b/Services/data/devicemodels/engine-01.json
@@ -53,10 +53,6 @@
             "Type": "javascript",
             "Path": "FirmwareUpdate-method.js"
         },
-        "Restart": {
-            "Type": "javascript",
-            "Path": "RestartEngine-method.js"
-        },
         "EmptyTank": {
             "Type": "javascript",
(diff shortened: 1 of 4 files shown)


Services/data/devicemodels/engine-01.json
```diff
@@ -53,10 +53,6 @@
             "Type": "javascript",
             "Path": "FirmwareUpdate-method.js"
         },
-        "Restart": {
```
ppathan: We are just hiding the functionality right now but not removing the actual file, right?

peterfelts: I removed the method file too, as part of my previous change #201. The method (before my change) was only a TODO anyway.

PR #209, opened 2018-05-02
Remove duplicate firmware-update related status keys in twin

peterfelts:
# Description and Motivation 
The FirmwareUpdate method reports the status of updates, which are written as reported properties in the twin (downloading, updating, updated, etc.). There were two properties being written to the twin: DeviceMethodStatus and FirmwareUpdateStatus. FirmwareUpdateStatus was always null. This change will result in DeviceMethodStatus not being reported to the twin. Only FirmwareUpdateStatus will be reported.
...

# Change type 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

### DIFF
--- a/Services/data/devicemodels/scripts/FirmwareUpdate-method.js
+++ b/Services/data/devicemodels/scripts/FirmwareUpdate-method.js
@@ -58,7 +58,7 @@ function main(context, previousState, previousProperties) {
     // Reboot - devices goes offline and comes online after 20 seconds
     log("Executing 'FirmwareUpdate' JavaScript method; Firmware version passed:" + context.Firmware);
 
-    var DeviceMethodStatusKey = "DeviceMethodStatus";
+    var DeviceMethodStatusKey = "FirmwareUpdateStatus";
     var FirmwareKey = "Firmware";
 
     // update the status to offline & firmware updating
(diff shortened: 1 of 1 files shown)


Services/data/devicemodels/scripts/FirmwareUpdate-method.js
```diff
@@ -58,7 +58,7 @@ function main(context, previousState, previousProperties) {
     // Reboot - devices goes offline and comes online after 20 seconds
     log("Executing 'FirmwareUpdate' JavaScript method; Firmware version passed:" + context.Firmware);
 
-    var DeviceMethodStatusKey = "DeviceMethodStatus";
+    var DeviceMethodStatusKey = "FirmwareUpdateStatus";
```
avanikp: Suggest renaming the var to FirmwareUpdateStatusKey

Services/data/devicemodels/scripts/FirmwareUpdate-method.js
```diff
@@ -58,7 +58,7 @@ function main(context, previousState, previousProperties) {
...
     var FirmwareKey = "Firmware";
```
avanikp: Also update DeviceMethodStatus in line 19 to FirmwareUpdateStatus

Services/data/devicemodels/scripts/FirmwareUpdate-method.js
peterfelts: This key shows up in RM, so it would probably make sense to keep the string as is.

PR #221, opened 2018-06-26
Merging Protobuf changes to private branch

hathind-ms:
# Description and Motivation 

...

# Change type 

- [ ] Bug fix
- [x ] New feature
- [ ] Enhancement

**Checklist:**

- [ ] All tests passed


Services/DeviceClient.cs
```diff
@@ -216,6 +217,73 @@ public class DeviceClient : IDeviceClient
             }
         }
 
+        private async Task SendProtoMessageAsync(string message, DeviceModel.DeviceModelMessageSchema schema)
```
dluc: nit: rename the method to `SendProtobufMessageAsync` (i.e. add the "buf" bit)

Services/DeviceClient.cs
```diff
@@ -30,6 +33,7 @@ public class DeviceClient : IDeviceClient
 
         // See also https://github.com/Azure/toketi-iothubreact/blob/master/src/main/scala/com/microsoft/azure/iot/iothubreact/MessageFromDevice.scala
         private const string CREATION_TIME_PROPERTY = "$$CreationTimeUtc";
+        private const string CLASSNAME_PROPERTY = "$$ClassName";
```
dluc: this is not used, is it? perhaps the method `SendProtoMessageAsync` was supposed to add a message property with this name?

hathind-ms: It is used in line 238

PR #223, opened 2018-07-01
2018 06 27 s3 perf

dluc:
# Description and Motivation 

1. Reduce time to start by using a common device connection string
2. Move some hard coded setting to config file, e.g. SDK timeouts and number of threads
3. Upgrade IoT SDK
4. Sync stock device models to use AMQP
5. Improve logging and docs

# Change type 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

**Checklist:**

- [x] All tests passed


Services/Devices.cs
```diff
@@ -134,78 +140,82 @@ public IDeviceClient GetClient(Device device, IoTHubProtocol protocol, IScriptIn
...
-        /// <summary>
-        /// Register the new device
-        /// </summary>
+        // Register a new device
         public async Task<Device> CreateAsync(string deviceId)
         {
-            this.SetupHub();
-            var now = DateTimeOffset.UtcNow;
+            this.CheckSetup();
+            var start = DateTimeOffset.UtcNow;
```
hathind-ms: for consistency sake, can we change this to start = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds here and then just use start var below in exception block

Services/Devices.cs
```diff
@@ -290,55 +312,69 @@ public IDeviceClient GetClient(Device device, IoTHubProtocol protocol, IScriptIn
...
-            this.registryCount++;
+            return result;
+        }
 
-            return this.registry;
+        private void CheckSetup()
+        {
+            if (this.setupDone) return;
+            throw new ApplicationException(this.GetType().FullName + " Setup incomplete. " +
+                                           "Call SetCurrentIotHubAsync() before using the instance.");
```
avanikp: Should  be SetCurrentIotHub() and not SetCurrentIotHubAsync()

Services/Devices.cs
```diff
@@ -74,54 +61,73 @@ public class Devices : IDevices
...
             this.twinReadsWritesEnabled = config.TwinReadWriteEnabled;
             this.registryCount = -1;
             this.setupDone = false;
         }
 
-        /// <summary>
-        /// Get IoTHub connection string from either the user provided value or the configuration
-        /// </summary>
+        // Get IoTHub connection string from either the user provided value or the configuration
         public void SetCurrentIotHub()
```
avanikp: How will this be called ? Can we consider calling this to setup iothub from constructor once ?

Services/Devices.cs
```diff
@@ -74,54 +61,73 @@ public class Devices : IDevices
...
             this.log = logger;
             this.twinReadsWritesEnabled = config.TwinReadWriteEnabled;
             this.registryCount = -1;
             this.setupDone = false;
         }
 
-        /// <summary>
-        /// Get IoTHub connection string from either the user provided value or the configuration
-        /// </summary>
+        // Get IoTHub connection string from either the user provided value or the configuration
```
avanikp: nit: comment needs to be updated

Services/DeviceClient.cs
```diff
@@ -124,23 +137,30 @@ public class DeviceClient : IDeviceClient
         /// </summary>
         public async Task UpdatePropertiesAsync(ISmartDictionary properties)
         {
+            // Temp code to detect where the "The given key was not present in the dictionary" exception is coming from
```
avanikp: temp code.... remove ?

SimulationAgent/SimulationRunner.cs
```diff
@@ -188,13 +194,18 @@ public void Start(Services.Models.Simulation simulation)
...
+                        var deviceModel = this.GetDeviceModel(model.Id, model.Override);
 
                         for (var i = 0; i < model.Count; i++)
                         {
                             this.CreateActorsForDevice(deviceModel, i, total);
                         }
                     }
+                    catch (ResourceNotFoundException)
+                    {
+                        this.IncreamentSimulationErrorsCount();
```
hathind-ms: Typo: Please change IncreamentSimulationErrorsCount -> IncrementSimulationErrorsCount

SimulationAgent/SimulationRunner.cs
```diff
@@ -213,7 +224,7 @@ public void Start(Services.Models.Simulation simulation)
 
                 this.TryToStartConnectionThread();
 
-                this.TryToStartTelemetryThread();
+                this.TryToStartTelemetryThreads();
```
hathind-ms: Indentation seems off

Services/Devices.cs
dluc: It's called when the simulation runner starts. The setup cannot be executed in the ctor because it has external dependencies and could bring down the application in case of exceptions. It's also going to become an async method

WebService/Runtime/Config.cs
```diff
@@ -186,6 +192,7 @@ private static IServicesConfig GetServicesConfig(IConfigData configData)
                 DeviceModelsScriptsFolder = MapRelativePath(configData.GetString(DEVICE_MODELS_SCRIPTS_FOLDER_KEY)),
                 IoTHubDataFolder = MapRelativePath(configData.GetString(IOTHUB_DATA_FOLDER_KEY)),
                 IoTHubConnString = connstring,
+                IoTHubSdkDeviceClientTimeout = configData.GetUInt(IOTHUB_SDK_DEVICE_CLIENT_TIMEOUT_KEY, 240000),
```
avanikp: 240000 should be const

SimulationAgent/SimulationRunner.cs
```diff
@@ -322,84 +333,137 @@ private void UpdateDevicesStateThread()
...
+            var tasks = new List<Task>();
+
             while (this.running)
             {
                 this.connectionLoopSettings.NewLoop();
                 var before = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
                 foreach (var device in this.deviceConnectionActors)
                 {
-                    device.Value.Run();
+                    tasks.Add(device.Value.RunAsync());
```
avanikp: This is repeated in multiple methods.. the code can be refactored to have common helper methods.

SimulationAgent/DeviceConnection/Connect.cs
```diff
@@ -41,23 +42,30 @@ public void Setup(IDeviceConnectionActor context, string deviceId, DeviceModel d
...
 
-                var timeSpent = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - now;
-                this.log.Debug("Device connected", () => new { this.deviceId, timeSpent });
+                var timeSpentMsecs = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - start;
+                this.log.Debug("Device connected", () => new { timeSpentMsecs, this.deviceId });
                 this.context.HandleEvent(DeviceConnectionActor.ActorEvents.Connected);
             }
+            catch (DeviceAuthFailedException e)
+            {
+                var timeSpentMsecs = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - start;
```
avanikp: nit: these can also be refactored to have common code.

Services/Devices.cs
```diff
@@ -74,54 +61,73 @@ public class Devices : IDevices
...
-            this.registry = this.registry.CreateFromConnectionString(connString);
-            this.ioTHubHostName = IotHubConnectionStringBuilder.Create(connString).HostName;
-            this.log.Info("Selected active IoT Hub for devices", () => new { this.ioTHubHostName });
+            try
+            {
+                // Retrieve connection string from file/storage
+                this.connString = this.connectionStringManager.GetIotHubConnectionString();
+
+                // Parse connection string, this triggers an exception if the string is invalid
+                IotHubConnectionStringBuilder cs = IotHubConnectionStringBuilder.Create(this.connString);
```
avanikp: nit: pls consider renaming 'cs' to something more descriptive.

SimulationAgent.Test/DeviceConnection/DeviceConnectionActorTest.cs
```diff
@@ -34,7 +35,10 @@ public DeviceConnectionActorTest()
             this.scriptInterpreter = new Mock<IScriptInterpreter>();
             this.devices = new Mock<IDevices>();
             this.rateLimiting = new Mock<IRateLimiting>();
-            this.fetchLogic = new Mock<Fetch>(
+            this.credentialsSetupLogic = new Mock<CredentialsSetup>(
+                this.devices.Object,
+                this.logger.Object);
+            this.fetchLogic = new Mock<FetchFromRegistry>(
```
avanikp: Missing tests for credentialsSetup, register cases.

Services/DeviceClient.cs
```diff
@@ -65,10 +68,21 @@ public class DeviceClient : IDeviceClient
         {
             if (this.client != null && !this.connected)
             {
-                // TODO: HTTP clients don't "connect", find out how HTTP connections are measured and throttled
-                //       https://github.com/Azure/device-simulation-dotnet/issues/85
-                await this.client.OpenAsync();
-                this.connected = true;
+                try
```
avanikp: missing unit tests for these as well

Services/data/devicemodels/chiller-02.json
```diff
@@ -4,7 +4,7 @@
     "Version": "0.0.1",
     "Name": "Faulty Chiller",
     "Description": "Faulty chiller with wrong pressure sensor. Pressure too high.",
-    "Protocol": "MQTT",
+    "Protocol": "AMQP",
```
avanikp: Why is this changed to AMQP for only some of the device models and not all ?

dluc: the others are already AMQP

SimulationAgent/SimulationRunner.cs
dluc: the code is repeated in 3 places, with one slightly different. This code is going to change considerably soon, and in my opinion the additional abstraction to deduplicate the code would make the code less readable and harder to refactor, so for the time being I will leave it as is.

SimulationAgent/DeviceConnection/Connect.cs
dluc: if we refactor this code out, the log would be generated on a different line of code, and we would lose the ability to see the real line number where the exception is generated

SimulationAgent.Test/DeviceConnection/DeviceConnectionActorTest.cs
dluc: I don't understand this comment, can you elaborate?

avanikp: I was referring to CredentialsSetup.cs, Register.cs, etc classes - unit tests for these classes are missing.

hathind-ms: We will add those tests as part of 'Unit Test Improvement' task.

PR #234, opened 2018-08-10
Multiple simulation support and adding stats to simulation definition

hathind-ms:
# Description and Motivation 
Adding support for multiple simulations while only 1 can be running
Moving Statistics to Simulation Definition

Example payload:
 ```json
 {
            "ETag": "\"00004d0c-0000-0000-0000-5b888db80000\"",
            "Id": "46f6ebeb-53fb-43ff-afb8-b1b686a123ac",
            "Name": "Truck",
            "Description": "",
            "Enabled": false,
            "Running": false,
            "IoTHubs": [
                {
                    "ConnectionString": "default",
                    "PreprovisionedIoTHubInUse": false,
                    "PreprovisionedIoTHubMetricsUrl": null
                }
            ],
            "StartTime": "2018-08-31T00:36:46+00:00",
            "EndTime": "2018-09-01T00:19:00+00:00",
            "StoppedTime": "2018-08-31T00:37:12+00:00",
            "DeviceModels": [
                {
                    "Id": "truck-01",
                    "Count": 6,
                    "Override": {
                        "Simulation": {
                            "Interval": "00:00:10"
                        },
                        "Telemetry": [
                            {
                                "Interval": "00:00:10"
                            }
                        ]
                    }
                }
            ],
            "Statistics": {
                "TotalMessagesSent": 12,
                "AverageMessagesPerSecond": 96.77,
                "FailedMessagesCount": 0,
                "ActiveDevicesCount": 0,
                "FailedDeviceConnectionsCount": 0,
                "FailedDeviceTwinUpdatesCount": 0,
                "SimulationErrorsCount": 0
            },
            "$metadata": {
                "$type": "Simulation;1",
                "$uri": "/v1/simulations/46f6ebeb-53fb-43ff-afb8-b1b686a123ac",
                "$created": "2018-08-31T00:36:46+00:00",
                "$modified": "2018-08-31T00:37:12+00:00"
            }
        }

```

...

# Change type 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/234)

### CODE REVIEW

File: WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs [hunk 1]
```diff
@@ -54,13 +69,17 @@ public class SimulationApiModel
         public SimulationApiModel()
         {
             this.Id = string.Empty;
+            this.Name = string.Empty;
```
avanikp: also add this.Description = string.Empty

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -38,13 +44,22 @@ public class SimulationApiModel
         [JsonProperty(PropertyName = "EndTime", NullValueHandling = NullValueHandling.Ignore)]
         public string EndTime { get; set; }
 
+        [JsonProperty(PropertyName = "StopTime", NullValueHandling = NullValueHandling.Ignore)]
+        public string StopTime { get; set; }
+
         [JsonProperty(PropertyName = "DeviceModels")]
         public IList<SimulationDeviceModelRef> DeviceModels { get; set; }
 
+        [JsonProperty(PropertyName = "TotalMsgs", NullValueHandling = NullValueHandling.Ignore)]
```
avanikp: nit: suggest renaming TotalMsgs to TotalMessages

Services/Simulations.cs
```diff
@@ -100,38 +100,53 @@ public async Task<IList<Models.Simulation>> GetListAsync()
...
+
+            return null;
         }
 
         /// <summary>
         /// Create a simulation.
         /// </summary>
         public async Task<Models.Simulation> InsertAsync(Models.Simulation simulation, string template = "")
         {
+            var usingDefaultTemplate = !string.IsNullOrEmpty(template) && template.ToLowerInvariant() == "default";
```
avanikp: make "default" a const

Services/Simulations.cs
```diff
@@ -100,38 +100,53 @@ public async Task<IList<Models.Simulation>> GetListAsync()
...
+
             // TODO: complete validation
             if (!string.IsNullOrEmpty(template) && template.ToLowerInvariant() != "default")
```
avanikp: change if condition to if (usingDefaultTemplate)

Services/Simulations.cs
```diff
@@ -100,38 +100,53 @@ public async Task<IList<Models.Simulation>> GetListAsync()
...
         }
 
         /// <summary>
         /// Create a simulation.
         /// </summary>
         public async Task<Models.Simulation> InsertAsync(Models.Simulation simulation, string template = "")
         {
+            var usingDefaultTemplate = !string.IsNullOrEmpty(template) && template.ToLowerInvariant() == "default";
+
             // TODO: complete validation
```
avanikp: remove TODO ?

Services/Simulations.cs
hathind-ms: This condition is not the same as usingDefaultTemplate.  There is a suttle difference, we want to exclude the case where template is not provided.

Services/Simulations.cs
hathind-ms: I didn't add this, so not sure what is meant by complete validation. I will keep it for now in case somebody needs to track it

Services/Simulations.cs

WebService/v1/Controllers/StatusController.cs
```diff
@@ -70,15 +70,15 @@ public sealed class StatusController : Controller
         }
 
         // TODO: reduce method complexity, refactor out some logic
-        [HttpGet]
-        public async Task<StatusApiModel> GetAsync()
+        [HttpGet("{id}")]
+        public async Task<StatusApiModel> GetAsync(string id)
```
dluc: I think we should drop simulation data from the /status endpoint. Indeed I suggested Hugh to move simulation stats from here to GET /simulation/<id>

Unless you already hooked up the UI to pass the ID, I would remove this.  Only useful info we could leave here is the number of running simulations, in future.

By the way, how will the UI know the simulation ID? I assume it will get all and use the first from the list? we might want to discuss this a bit

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -26,6 +26,12 @@ public class SimulationApiModel
         [JsonProperty(PropertyName = "Id")]
         public string Id { get; set; }
 
+        [JsonProperty(PropertyName = "Name")]
+        public string Name { get; set; }
+
+        [JsonProperty(PropertyName = "Desc", NullValueHandling = NullValueHandling.Ignore)]
```
dluc: I think you can drop the NullValueHandling, and let the API returng a null field. That will help a user willing to edit the description, changing it from null to something else

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -38,13 +44,22 @@ public class SimulationApiModel
         [JsonProperty(PropertyName = "EndTime", NullValueHandling = NullValueHandling.Ignore)]
         public string EndTime { get; set; }
 
+        [JsonProperty(PropertyName = "StopTime", NullValueHandling = NullValueHandling.Ignore)]
+        public string StopTime { get; set; }
```
dluc: when creating a simulation, will we pass the EndTime or StopTime?  Could you update the documentation in the repo to describe how these new fields are used, e.g. when creating/editing a simulation?

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -38,13 +44,22 @@ public class SimulationApiModel
...
+
         [JsonProperty(PropertyName = "DeviceModels")]
         public IList<SimulationDeviceModelRef> DeviceModels { get; set; }
 
+        [JsonProperty(PropertyName = "TotalMsgs", NullValueHandling = NullValueHandling.Ignore)]
+        public int TotalMessagesSent { get; set; }
+
+        [JsonProperty(PropertyName = "AvgMsgs", NullValueHandling = NullValueHandling.Ignore)]
+        public double AverageMessagesSent { get; set; }
```
dluc: I think the scope of the PR is growing from multiple simulations to something else?  also, the average should have a reference, e.g. is this per second/per minute?  I'd like to discuss the new API before merging this PR, to see how it looks like and what's the expected behavior in edge cases. For instance, it's not clear which fields are readonly (statistics) and which not. A client won't know which fields are required when creating a simulation, and which ones are not. As a side note, I was thinking to move all the statistics into an object, something like:
```json
{
  ...
  "Statistics": {
    "TotalDevices": 20000,
    "DevicesCurrentlyConnected": 15200,
    "AverageMessagesPerSecond": 2200,
    ...
  }
...
```

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -38,13 +44,22 @@ public class SimulationApiModel
...
+
         [JsonProperty(PropertyName = "$metadata", Order = 1000)]
         public IDictionary<string, string> Metadata => new Dictionary<string, string>
         {
-            { "$type", "Simulation;" + Version.NUMBER },
+            { "$type", "Simulation" },
```
dluc: we should leave the version number, this is the model version that the UI can use to understand the schema of the response - it's the same approach across all our APIs

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
hathind-ms: Again, this is interim solution to store these values and will change with Hugh's work. For now I will move statistics under one object. 

Let' s chat tomorrow

Services/Devices.cs
```diff
@@ -313,10 +315,12 @@ public async Task DeleteListAsync(IEnumerable<string> deviceIds)
...
 
-        // Generate a device Id
-        public string GenerateId(string deviceModelId, int position)
+        /// <summary>
+        /// Generate a device Id
+        /// </summary>
+        public string GenerateId(string simulationId, string deviceModelId, int position)
         {
-            return deviceModelId + "." + position;
+            return simulationId + "." + deviceModelId + "." + position;
```
dluc: we'll do this in the scale work, so it's ok to remove it for now

Services/Models/Simulation.cs
```diff
@@ -11,18 +11,20 @@ public class Simulation
     {
         private DateTimeOffset? startTime;
         private DateTimeOffset? endTime;
+        private DateTimeOffset? stopTime;
         private string iotHubConnectionString;
 
         [JsonIgnore]
         public string ETag { get; set; }
 
-        [JsonIgnore]
```
dluc: as discussed offline, undo this change. Add comment explaining why we use JsonIgnore, smth like:  "when written to storage, Id and Etag are not serialized, and are instead written in dedicated columns..."

Services/Models/Simulation.cs
```diff
@@ -36,6 +38,12 @@ public class Simulation
             set => this.endTime = value ?? DateTimeOffset.MaxValue;
         }
 
+        public DateTimeOffset? StopTime
```
dluc: pls add comment describing the use case/difference between EndTime and StopTime

Services/Simulations.cs
```diff
@@ -91,7 +92,7 @@ public async Task<IList<Models.Simulation>> GetListAsync()
                 result.Add(simulation);
             }
 
-            return result;
+            return result.OrderByDescending(s => s.Modified).ToList();
```
dluc: order by by creation

dluc: TODO: eventually this will return a long list, potentially paginated, so we need to discuss how to get the latest ones first, or paginate etc.

Services/Simulations.cs
```diff
@@ -169,31 +185,34 @@ public async Task<Models.Simulation> InsertAsync(Models.Simulation simulation, s
...
+            if (existingSimulation != null)
             {
                 this.log.Info("Modifying simulation");
 
                 if (simulation.ETag == "*")
                 {
-                    simulation.ETag = simulations[0].ETag;
-                    this.log.Warn("The client used ETag='*' choosing to overwrite the current simulation");
+                    simulation.ETag = existingSimulation.ETag;
+                    this.log.Info("The client used ETag='*' choosing to overwrite the current simulation");
```
dluc: pls make sure this is a Warning, because clients should never use "*" (that's a development feature only)

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -74,12 +90,16 @@ public Simulation ToServiceModel(string id = "")
             {
                 ETag = this.ETag,
                 Id = this.Id,
+                Name = this.Name,
+                Description = this.Description,
                 // When unspecified, a simulation is enabled
                 Enabled = this.Enabled ?? true,
                 StartTime = DateHelper.ParseDateExpression(this.StartTime, now),
                 EndTime = DateHelper.ParseDateExpression(this.EndTime, now),
+                StopTime = DateHelper.ParseDateExpression(this.StopTime, now),
```
dluc: the client should not be allowed to change this field (readonly field)

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -74,12 +90,16 @@ public Simulation ToServiceModel(string id = "")
...
                 IotHubConnectionString = SimulationIotHub.ToServiceModel(this.IotHub),
-                DeviceModels = this.DeviceModels?.Select(x => x.ToServiceModel()).ToList()
+                DeviceModels = this.DeviceModels?.Select(x => x.ToServiceModel()).ToList(),
+                Statistics = SimulationStatisticsRef.ToServiceModel(this.Statistics)
```
dluc: should be readonly, i.e. a web client cannot modify the stats

Services/Simulations.cs
```diff
@@ -100,38 +109,53 @@ public async Task<IList<Models.Simulation>> GetListAsync()
...
-                this.log.Warn("There is already a simulation");
+                this.log.Warn("There is already a running simulation");
                 throw new ConflictingResourceException(
-                    "There is already a simulation. Only one simulation can be created.");
+                    "There is already a simulation. Simulation cannot be created while another simulation is running.");
             }
 
             // Note: forcing the ID because only one simulation can be created
-            simulation.Id = SIMULATION_ID;
+            simulation.Id = usingDefaultTemplate ? DEFAULT_SIMULATION_ID : Guid.NewGuid().ToString();
```
peterfelts: It might be good to specify the format of the GUID to use here, something like: 
Guid.NewGuid().ToString("N").ToUpperInvariant();

This is not a must-have, but perhaps it could avoid errors in the future if other code is expecting the simulation ID to be using a different format.

SimulationAgent/Agent.cs
```diff
@@ -136,9 +215,16 @@ private void CheckForNewSimulation(Services.Models.Simulation newSimulation)
...
+            if (this.simulation != null && this.simulation.Enabled && !this.simulation.ShouldBeRunning())
             {
+                this.simulation.Enabled = false;
+                this.simulation.Statistics.AverageMessagesPerSecond = this.rateReporter.GetThroughputForMessages();
+                this.simulation.Statistics.TotalMessagesSent = this.runner.TotalMessagesCount;
+
                 this.runner.Stop();
+
+                // Update simulation statistics in storage
+                this.simulations.UpdateStorage(this.simulation);
```
dluc: should be async

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApiModel
+{
+    public class SimulationStatistics
+    {
+        [JsonProperty(PropertyName = "TotalMessagesSent")]
+        public long TotalMessagesSent { get; set; }
+
+        [JsonProperty(PropertyName = "AverageMessagesPerSecond")]
+        public double AverageMessagesPerSecond { get; set; }
```
dluc: note, we've discussed this and the plan is to keep these stats for the last N minutes, so this value won't be a global average for the entire duration of the simulation. I'm wondering if we should make that explicit with a name like `CurrentAverageMessagesPerSecond` or similar - thoughts?

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        [JsonProperty(PropertyName = "FailedMessagesCount")]
+        public long FailedMessagesCount { get; set; }
+
+        [JsonProperty(PropertyName = "ActiveDevicesCount")]
+        public long ActiveDevicesCount { get; set; }
```
dluc: could you comments on all the fields, helping to understand what to expect? For instance what do we mean by "Active" ?

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
```diff
@@ -0,0 +1,71 @@
...
+            this.FailedMessagesCount = 0;
+            this.FailedDeviceConnectionsCount = 0;
+            this.FailedDeviceTwinUpdatesCount = 0;
+            this.SimulationErrorsCount = 0;
+        }
+
+        // Map API model to service model
+        public static Services.Models.Simulation.StatisticsRef ToServiceModel(SimulationStatistics statistics)
+        {
+            if(statistics != null)
```
dluc: spacing

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
```
dluc: nit: invert if to reduce indentation

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
```diff
@@ -0,0 +1,71 @@
...
+                return new Services.Models.Simulation.StatisticsRef {
```
dluc: we'll have to revisit the class names, e.g. the idea about classes with a "Ref" suffix was to be a subset of the complete class, i.e. a reference to the actual object. I think this is the second instance where that idea is not applied, so at this point I'm not sure what "Ref" means anymore :-)  We'll see this separately once we're done with major tasks, not a blocker because it doesn't impact clients

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -18,34 +21,69 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
...
             this.log = logger;
         }
 
         [HttpGet]
         public async Task<SimulationListApiModel> GetAsync()
         {
-            return new SimulationListApiModel(await this.simulationsService.GetListAsync());
+            var simulationList = await this.simulationsService.GetListAsync();
+            var simulationListApiModel = new SimulationListApiModel();
+            foreach (var x in simulationList)
```
dluc: could all this work be done in the `SimulationListApiModel` class?

dluc: looks like all the work done here is mapping from one layer to another, mapping tasks that usually have in the model classes

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -18,34 +21,69 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
...
+            return simulationListApiModel;
         }
 
         [HttpGet("{id}")]
         public async Task<SimulationApiModel> GetAsync(string id)
         {
-            return SimulationApiModel.FromServiceModel(await this.simulationsService.GetAsync(id));
+            var simulation = await this.simulationsService.GetAsync(id);
+
+            if (simulation == null)
```
dluc: doesn't line 75 throw a ResourceNotFoundException?

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -18,34 +21,69 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
...
+            {
+                this.log.Warn("Simulation not found", () => new { id });
+                throw new BadRequestException("No data or invalid id provided.");
```
dluc: should be `ResourceNotFoundException` resulting in a 404 Not Found.  This instead would lead to 400 Bad Request, which we use only when a client is broken, i.e. it has a bug and is making an invalid request

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
hathind-ms: These represent final state as ie when Simulation ends, we save TotalMessagesSent and AverageMessagesPerSecond in simulation def.

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
hathind-ms: This StatisticsRef is a subclass, it only contains 2 properties instead of all 5

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -94,10 +120,22 @@ public static SimulationApiModel FromServiceModel(Simulation value)
...
+                Running = value.ShouldBeRunning(),
+                StartTime = value.StartTime.ToString(),
+                EndTime = value.EndTime.ToString(),
+                StoppedTime = value.StoppedTime.ToString(),
+                IotHubs = new List<SimulationIotHub>()
             };
 
+            foreach (var iotHubConnectionString in value.IotHubConnectionStrings)
+            {
+                var iotHub = new SimulationIotHub { ConnectionString = iotHubConnectionString };
```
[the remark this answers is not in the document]
hathind-ms: Yes, I call connectionStringManager.RedactAndStoreAsync(iotHubConnectionString) before storing

WebService/v1/Models/SimulationApiModel/SimulationIotHub.cs
```diff
@@ -14,15 +14,34 @@ public class SimulationIotHub
...
         public string ConnectionString { get; set; }
 
+        [JsonProperty(PropertyName = "PreprovisionedIoTHub")]
+        public bool PreprovisionedIoTHub { get; set; }
+
+        [JsonProperty(PropertyName = "PreprovisionedIoTHubInUse")]
+        public bool PreprovisionedIoTHubInUse { get; set; }
+
+        [JsonProperty(PropertyName = "PreprovisionedIoTHubMetricsUrl")]
+        public string PreprovisionedIoTHubMetricsUrl { get; set; }
```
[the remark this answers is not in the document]
hathind-ms: done

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -18,74 +21,103 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
     public class SimulationsController : Controller
     {
         private const int MAX_DELETE_DEVICES = 100;
+
         private readonly ISimulations simulationsService;
+        private readonly IServicesConfig servicesConfig;
+        private readonly IDeploymentConfig deploymentConfig;
         private readonly IIotHubConnectionStringManager connectionStringManager;
+        private readonly IPreprovisionedIotHub preprovisionedIotHub;
```
dluc: doesn't seem to be used, is it?

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -18,74 +21,103 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
...
         [HttpGet]
         public async Task<SimulationListApiModel> GetAsync()
         {
-            return new SimulationListApiModel(await this.simulationsService.GetListAsync());
+            var simulationList = await this.simulationsService.GetListAsync();
+            var simulationListApiModel = new SimulationListApiModel();
+            foreach (var simulation in simulationList)
+            {
+                var simulationApiModel = SimulationApiModel.FromServiceModel(
+                    simulation, servicesConfig, deploymentConfig, connectionStringManager, simulationRunner, rateReporter);
```
dluc: code style, here and other parts, please add `this.`

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -165,7 +217,95 @@ public async Task ValidateInputRequest(ILogger log, IIotHubConnectionStringManag
...
+            // Return the URL only when the simulation is running with the configured conn string
+            if (csInUse != csInConf) return string.Empty;
+
+            return $"https://portal.azure.com/{deploymentConfig.AzureSubscriptionDomain}" +
+                   $"#resource/subscriptions/{deploymentConfig.AzureSubscriptionId}" +
+                   $"/resourceGroups/{deploymentConfig.AzureResourceGroup}" +
+                   $"/providers/Microsoft.Devices/IotHubs/{deploymentConfig.AzureIothubName}/Metrics";
+        }
+
+
```
dluc: remove empty lines

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                return null;
+            }
+
+            return new SimulationStatistics
+            {
+                TotalMessagesSent = statistics.TotalMessagesSent,
+                AverageMessagesPerSecond = Math.Round(statistics.AverageMessagesPerSecond, 2)
+            };
+
```
dluc: remove empty line (btw, in my IDE all these code style issues are highlighted in red, making it very easy to spot and fix them - if you haven't, you can configure VS and Rider to do so)

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -8,8 +8,8 @@ The service supports only one simulation, with ID "1".
 ### Creation with POST
 
 When invoking the API using the POST HTTP method, the service will always
-attempt to create a new simulation. Since the service allows
-to create only one simulation, retrying the request will result in errors
+attempt to create a new simulation. The service allows
+to create multiple simulation but there can only be one running simulation. Retrying the request to create a new running simulation will result in errors
 after the first request has been successfully completed.
```
dluc: I think we need to clarify this part. If it is possible to create multiple simulations, but there can be only one running, when will POST allow to create a second simulation and when not? Does the service check if there is already a simulation running?

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -205,6 +208,7 @@ Content-Type: application/json; charset=utf-8
 ```
 ```json
 {
+  "Name": "Sample Simulation",
```
dluc: There is also a Description field right?

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -466,35 +472,55 @@ Content-Type: application/json; charset=utf-8
...
+  "Description": "This is a sample simulation",
+  "Enabled": false,
+  "Running": false,
+  "IoTHubs": [
+      {
+          "ConnectionString": "default",
+          "PreprovisionedIoTHub": true,
+          "PreprovisionedIoTHubInUse": false,
+          "PreprovisionedIoTHubMetricsUrl": null
```
dluc: these 2 seems inconsistent. If the simulation is using the `default` connection string, then `PreprovisionedIoTHubInUse` should be true, and `PreprovisionedIoTHubInUse` should have a value

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -466,35 +472,55 @@ Content-Type: application/json; charset=utf-8
...
   "$metadata": {
-    "$type": "Simulation;1",
-    "$uri": "/v1/simulations/1",
-    "$version": "1",
-    "$created": "2017-05-31T00:47:18+00:00",
-    "$modified": "2017-05-31T00:47:18+00:00"
+      "$type": "Simulation;1",
+      "$uri": "/v1/simulations/1",
+      "$created": "2018-08-31T00:30:00+00:00",
+      "$modified": "2018-08-31T00:30:00+00:00"
```
dluc: given the example, this should match the stopped time, when the user stopped it

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -466,35 +472,55 @@ Content-Type: application/json; charset=utf-8
...
   }
 }
 ```
 
+There are some read-only properties in the api response. `Running` is a calculated property implying whether the simulation is currently running. The `StoppedTime` is the time when the simulation was manually stopped before it's scheduled `EndTime`. `PreprovisionedIoTHubMetricsUrl` implies whether the simulation is currently running using the `PreprovisionedIoTHub`. 'PreprovisionedIoTHubMetricsUrl' is the url to navigate to the IoT Hub metrics page.
```
dluc: split long line, so we can easily add comments in the PR and use IDE's split windows without using wordwrapping

docs/API_SPECS_SIMULATIONS.md
dluc: please also split long lines, around the 76th column

WebService/v1/Controllers/SimulationsController.cs
hathind-ms: good catch. it's not used

docs/API_SPECS_SIMULATIONS.md
hathind-ms: added

Services/Concurrency/RateLimiting.cs
```diff
@@ -97,6 +97,6 @@ public long GetPauseForNextMessage()
         /// <summary>
         /// Get message throughput (messages per second)
         /// </summary>
-        public double GetThroughputForMessages() => this.messaging.GetThroughputForMessages();
+        public double GetThroughputForMessages() => System.Math.Round(this.messaging.GetThroughputForMessages(), 2);
```
dluc: could you use Math.Ceiling instead?

* Math.Round(0.005, 2) => 0
* Math.Ceiling(0.005 * 100) / 100 => 0.01

docs/API_SPECS_SIMULATIONS.md
dluc: there are some errors in the text, here's a suggested edit:
```
* `Running` is a calculated property, reporting whether the simulation is
  currently running.
* `StoppedTime` is the time when the simulation was manually
  stopped before it's scheduled `EndTime`. 
* `PreprovisionedIoTHubInUse`: whether the simulation ran or is running
  using the Azure IoT Hub created during the solution deployment.
* 'PreprovisionedIoTHubMetricsUrl': the url to navigate to the IoT Hub
  metrics page. 
* `Statistics` is a set of read only properties providing counters
  at a given time. When the simulation is stopped, the current values
  for `TotalMessagesSent` and `AverageMessagesPerSecond` are retrieved from
  storage.
```

Services/Simulations.cs
```diff
@@ -107,38 +114,53 @@ public async Task<IList<Models.Simulation>> GetListAsync()
         public async Task<Models.Simulation> GetAsync(string id)
         {
             var item = await this.storage.GetAsync(STORAGE_COLLECTION, id);
-            var simulation = JsonConvert.DeserializeObject<Models.Simulation>(item.Data);
-            simulation.ETag = item.ETag;
-            return simulation;
+
+            if (item != null)
```
dluc: nit, invert if:

```
if (item == null) return null;
```

Services/Simulations.cs
```diff
@@ -107,38 +114,53 @@ public async Task<IList<Models.Simulation>> GetListAsync()
...
         /// <summary>
         /// Create a simulation.
         /// </summary>
         public async Task<Models.Simulation> InsertAsync(Models.Simulation simulation, string template = "")
         {
-            // TODO: complete validation
-            if (!string.IsNullOrEmpty(template) && template.ToLowerInvariant() != "default")
+            var usingDefaultTemplate = !string.IsNullOrEmpty(template) && template.ToLowerInvariant() == DEFAULT_TEMPLATE_NAME;
+
+            if (!string.IsNullOrEmpty(template) && template.ToLowerInvariant() != DEFAULT_TEMPLATE_NAME)
```
dluc: could use `usingDefaultTemplate` here?

Services/Simulations.cs
```diff
@@ -107,38 +114,53 @@ public async Task<IList<Models.Simulation>> GetListAsync()
...
             {
                 this.log.Warn("Unknown template name", () => new { template });
                 throw new InvalidInputException("Unknown template name. Try 'default'.");
             }
 
+            if (!usingDefaultTemplate && string.IsNullOrEmpty(simulation.Name))
```
dluc: when usingDefaultTemplate == true && simulation.Name is empty, we could generate a name, e.g. "Default Simulation 2018-12-01T14:55:00Z"

Services/Simulations.cs
```diff
@@ -107,38 +114,53 @@ public async Task<IList<Models.Simulation>> GetListAsync()
...
+                throw new InvalidInputException("Simulation name is required.");
+            }
+
             var simulations = await this.GetListAsync();
-            if (simulations.Count > 0)
+            var activeSimulation = simulations.Where(a => a.ShouldBeRunning());
+            if (activeSimulation.Any())
             {
-                this.log.Warn("There is already a simulation");
+                this.log.Warn("There is already a running simulation");
```
dluc: as per discussion, we're removing this, correct? so multiple simulations can be created, then the Agent temporarily will execute only one?

Services/Simulations.cs
dluc: it's a nit but I like the suggestion, only I'd prefer lowercase

Services/Simulations.cs
hathind-ms: removed

Services/Simulations.cs
```diff
@@ -156,16 +178,25 @@ public async Task<Models.Simulation> InsertAsync(Models.Simulation simulation, s
...
             //      https://github.com/Azure/device-simulation-dotnet/issues/129
-            simulation.IotHubConnectionString = await this.connectionStringManager.RedactAndStoreAsync(simulation.IotHubConnectionString);
+
+            var iotHubConnectionStrings = new List<string>(simulation.IotHubConnectionStrings);
+            simulation.IotHubConnectionStrings.Clear();
+            foreach (var iotHubConnectionString in iotHubConnectionStrings)
+            {
+                var connString = await this.connectionStringManager.RedactAndStoreAsync(iotHubConnectionString);
+                simulation.IotHubConnectionStrings.Add(connString);
+            }
```
dluc: nit: this block could be simplified using a for loop:
```
for (var index = 0; index < simulation.IotHubConnectionStrings.Count; index++)
{
    var connString = await this.connectionStringManager.RedactAndStoreAsync(simulation.IotHubConnectionStrings[index]);
    simulation.IotHubConnectionStrings[index] = connString;
}
```

Services/Simulations.cs
```diff
@@ -210,23 +240,27 @@ public async Task<Models.Simulation> UpsertAsync(Models.Simulation simulation)
...
             //      storage adapter would address this
             //      https://github.com/Azure/device-simulation-dotnet/issues/129
-            simulation.IotHubConnectionString = await this.connectionStringManager.RedactAndStoreAsync(simulation.IotHubConnectionString);
+            var iotHubConnectionStrings = new List<string>(simulation.IotHubConnectionStrings);
+            simulation.IotHubConnectionStrings.Clear();
+            foreach (var iotHubConnectionString in iotHubConnectionStrings)
+            {
+                var connString = await this.connectionStringManager.RedactAndStoreAsync(iotHubConnectionString);
+                simulation.IotHubConnectionStrings.Add(connString);
+            }
```
dluc: nit: same comment here, you might even extract a method out, or update the connectionStringManager to accept in input a list

Services/Simulations.cs
```diff
@@ -293,6 +332,18 @@ public async Task AddDeviceAsync(string id)
             await this.storage.CreateAsync(DEVICES_COLLECTION, id, id);
         }
 
+        /// <summary>
+        /// Update simulation definition in storage.
+        /// </summary>
+        public async Task UpdateStorage(Models.Simulation simulation)
```
dluc: missing `Async` suffix

SimulationAgent/Agent.cs
```diff
@@ -59,17 +63,27 @@ public async Task RunAsync()
                 {
                     this.log.Debug("------ Checking for simulation changes ------");
 
-                    var newSimulation = (await this.simulations.GetListAsync()).FirstOrDefault();
-                    this.log.Debug("Simulation loaded", () => new { newSimulation });
+                    var simulationList = await this.simulations.GetListAsync();
+
+                    // currently we support only 1 running simulation so the result should return only 1 item
```
dluc: please add to the comment: "The list of simulations returned is sorted by creation date, so the agent runs the first that was created"

SimulationAgent/Agent.cs
```diff
@@ -59,17 +63,27 @@ public async Task RunAsync()
...
+                    var runningSimulation = simulationList.FirstOrDefault(s => s.ShouldBeRunning());
+
+                    this.log.Debug("Simulation loaded", () => new { runningSimulation });
 
                     // if the simulation is removed from storage & we're running stop simulation.
-                    this.CheckForDeletedSimulation(newSimulation);
+                    var id = this.simulation?.Id;
+                    var prevSimulation = simulationList.FirstOrDefault(s => s.Id == id);
+                    this.CheckForDeletedSimulation(prevSimulation);
```
dluc: this code is going to be rewritten, however: if not too hard, move the added logic inside one of the 4 methods (CheckForDeletedSimulation, CheckForChangedSimulation, CheckForNewSimulation, etc.) or add a new one.  This method used to be very long, with a lot of cases to handle, and we split it into those 4 private methods.   Anyway, as I said, this will be rewritten, so consider it just an observation

hathind-ms: I will leave this as is for now.

Services/Concurrency/RateLimiting.cs
```diff
@@ -97,6 +97,6 @@ public long GetPauseForNextMessage()
         /// <summary>
         /// Get message throughput (messages per second)
         /// </summary>
-        public double GetThroughputForMessages() => System.Math.Round(this.messaging.GetThroughputForMessages(), 2);
+        public double GetThroughputForMessages() => System.Math.Ceiling(this.messaging.GetThroughputForMessages() * 100)/100;
```
dluc: spacing before/after "/"

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
```diff
@@ -46,35 +46,16 @@ public SimulationStatistics()
...
-            {
-                return null;
-            }
+            if (statistics == null) return null;
 
             return new SimulationStatistics
             {
                 TotalMessagesSent = statistics.TotalMessagesSent,
-                AverageMessagesPerSecond = Math.Round(statistics.AverageMessagesPerSecond, 2)
+                AverageMessagesPerSecond = Math.Ceiling(statistics.AverageMessagesPerSecond * 100)/100
```
dluc: spacing

SimulationAgent/Agent.cs
```diff
@@ -218,6 +218,7 @@ private async Task CheckForStopOrStartToSimulationAsync()
             if (this.simulation != null && this.simulation.Enabled && !this.simulation.ShouldBeRunning())
             {
                 this.simulation.Enabled = false;
+                this.simulation.Modified = DateTimeOffset.Now;
```
dluc: this could be done when calling simulations.UpdateAsync() so we're sure it will always happen and callers don't need to worry

dluc: btw, this should be `UctNow`

Services/Simulations.cs
```diff
@@ -293,6 +319,18 @@ public async Task AddDeviceAsync(string id)
...
```
dluc: could we use `UpsertAsync` instead of this method? looks like they do the same thing, no? "Upsert" is used for creation and update, so it should work

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -18,74 +21,101 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
...
+            var simulationList = await this.simulationsService.GetListAsync();
+            var simulationListApiModel = new SimulationListApiModel();
+            foreach (var simulation in simulationList)
+            {
+                var simulationApiModel = SimulationApiModel.FromServiceModel(
+                    simulation, this.servicesConfig, this.deploymentConfig, this.connectionStringManager, this.simulationRunner, this.rateReporter);
+                simulationListApiModel.Items.Add(simulationApiModel);
+            }
+
+            return simulationListApiModel;
```
dluc: could use the same approach used in DeviceModelListApiModel, e.g. `return SimulationListApiModel.FromServiceModel(simulationList)`

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -133,8 +163,20 @@ public async Task<SimulationApiModel> GetAsync(string id)
                 throw new BadRequestException("No data or invalid data provided");
             }
 
+            var patchServiceModel = patch.ToServiceModel(id);
+
+            if (patchServiceModel.Enabled == false)
+            {
+                patchServiceModel.Statistics = new Services.Models.SimulationStatistics
```
dluc: I would avoid doing this:
* the Agent already stores the stats when stopping the simulation
* if there are multiple simulation running (in future) there are multiple `rateReporter` objects, one per simulation

WebService/v1/Controllers/StatusController.cs
```diff
@@ -27,50 +24,29 @@ public sealed class StatusController : Controller
...
             this.servicesConfig = servicesConfig;
-            this.deploymentConfig = deploymentConfig;
-            this.connectionStringManager = connectionStringManager;
-            this.simulationRunner = simulationRunner;
-            this.rateReporter = rateLimiting;
         }
 
         // TODO: reduce method complexity, refactor out some logic
-        [HttpGet]
+        [HttpGet()]
```
dluc: needed?

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -165,7 +220,92 @@ public async Task ValidateInputRequest(ILogger log, IIotHubConnectionStringManag
...
+                {
+                    iotHub.PreprovisionedIoTHubInUse = this.IsPreprovisionedIoTHubInUse(servicesConfig, connectionStringManager);
+                    var preprovisionedIoTHubMetricsUrl = this.GetIoTHubMetricsUrl(servicesConfig, deploymentConfig, connectionStringManager);
+                    iotHub.PreprovisionedIoTHubMetricsUrl = preprovisionedIoTHubMetricsUrl;
+                }
+            }
+
+            if (isRunning)
+            {
+                // Average messages per second count
```
dluc: nit: "Average messages per second count" => "Average messages per second frequency in the last minutes"

WebService/v1/Models/SimulationApiModel/SimulationIotHub.cs
```diff
@@ -14,15 +14,28 @@ public class SimulationIotHub
         [JsonProperty(PropertyName = "ConnectionString")]
         public string ConnectionString { get; set; }
 
+        [JsonProperty(PropertyName = "PreprovisionedIoTHubInUse")]
+        public bool PreprovisionedIoTHubInUse { get; set; }
+
+        [JsonProperty(PropertyName = "PreprovisionedIoTHubMetricsUrl")]
+        public string PreprovisionedIoTHubMetricsUrl { get; set; }
```
dluc: note: throughout the code we have ["hub preprovisioned" "preprovisioned hub"] and ["preprovisioned iot hub"]. It would be nice to align to one of the two, i.e. with or without the "iot" word

WebService/v1/Models/SimulationListApiModel.cs
```diff
@@ -21,12 +21,5 @@ public SimulationListApiModel()
         {
             this.Items = new List<SimulationApiModel.SimulationApiModel>();
         }
-
-        /// <summary>Map a service model to the corresponding API model</summary>
-        public SimulationListApiModel(IEnumerable<Services.Models.Simulation> simulations)
```
dluc: I'd prefer this approach to having this logic in the controller, it's part of the mapping logic

WebService/v1/Models/SimulationApiModel/SimulationIotHub.cs
hathind-ms: We will address this is a separate PR. Already scope of this PR has increased a lot

WebService/v1/Controllers/SimulationsController.cs
hathind-ms: I tried but that does not work. When a simulation is manually stopped the Stop/Update methods in Agent are not triggered

WebService/v1/Models/SimulationListApiModel.cs
dluc: :white_check_mark:

WebService/v1/Models/SimulationApiModel/SimulationIotHub.cs
dluc: :white_check_mark:

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
dluc: :white_check_mark:

WebService/v1/Controllers/StatusController.cs
dluc: :white_check_mark:

WebService/v1/Controllers/SimulationsController.cs
dluc: :white_check_mark:

WebService/v1/Controllers/SimulationsController.cs
dluc: :white_check_mark:

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -1,10 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+using System.Collections.Generic;
```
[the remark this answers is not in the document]
dluc: :white_check_mark:

Services/Simulations.cs
dluc: :white_check_mark:

SimulationAgent/Agent.cs
dluc: :white_check_mark:

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
dluc: :white_check_mark:

SimulationAgent/Agent.cs
dluc: :white_check_mark:

Services/Simulations.cs
dluc: :white_check_mark:

PR #236, opened 2018-08-16
Webinsights - Sending Diagnostics Data from Microservice to Diagnostics Backend

sasathy:
Logging start simulation, stop simulation and service errors to diagnostics backend.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/236)


Services/Diagnostics/SendDataToDiagnostics.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Net;
+using System.Text;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics
+{
+    public interface ISendDataToDiagnostics
+    {
+        Task<IHttpResponse> SendDiagnosticsData(string eventType, string message);
+    }
+    public class SendDataToDiagnostics: ISendDataToDiagnostics
```
avanikp: Suggest renaming this class to LogDiagnostics or something smilar

Services.Test/SendingDataToDiagnostics.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.StorageAdapter;
+using Moq;
+using Newtonsoft.Json;
+using Services.Test.helpers;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Services.Test
+{
+    public class SendingDataToDiagnostics
```
avanikp: class name should be classname + "Test"

Services/Diagnostics/SendDataToDiagnostics.cs
```diff
@@ -0,0 +1,48 @@
+﻿using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Http;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Runtime;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
```
avanikp: nit: sort

Services.Test/SendingDataToDiagnostics.cs
```diff
@@ -0,0 +1,41 @@
+﻿
```
avanikp: missing copyright

Services/Diagnostics/SendDataToDiagnostics.cs
```diff
@@ -0,0 +1,48 @@
+﻿using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Http;
```
avanikp: missing copyright

SimulationAgent/SimulationRunner.cs
```diff
@@ -274,6 +299,16 @@ public void Stop()
             }
         }
 
+        private HttpRequest PrepareRequest(string path, string content = null)
```
avanikp: This should be removed, it's not used here

SimulationAgent/SimulationRunner.cs
```diff
@@ -143,12 +154,26 @@ public class SimulationRunner : ISimulationRunner
...
         /// sending the telemetry.
         /// </summary>
-        public void Start(Services.Models.Simulation simulation)
+        public async void Start(Services.Models.Simulation simulation)
         {
             // Use `starting` to exit as soon as possible, to minimize the number 
             // of threads pending on the lock statement
-            if (this.starting || this.running) return;
+            if (this.starting || this.running)
+            {
```
avanikp: this code should be moved to Agent.cs instead of here.

SimulationAgent/SimulationRunner.cs
```diff
@@ -143,12 +154,26 @@ public class SimulationRunner : ISimulationRunner
...
+                DateTime now = DateTime.Now;
+                TimeSpan duration = now - this.lastPolledTime;
+                
 
+                // Send heartbeat every 24 hours
+                if (duration.Days >= 1)
+                {
+                    this.lastPolledTime = DateTime.Now;
+                    await this.sendDataToDiagnostics.SendDiagnosticsData("Service Heartbeat","");
```
avanikp: ServiceHeartbeat

SimulationAgent/SimulationRunner.cs
```diff
@@ -143,12 +154,26 @@ public class SimulationRunner : ISimulationRunner
...
+                }
+                return;
+            }
+            await this.sendDataToDiagnostics.SendDiagnosticsData("Service Strt", "");
```
avanikp: ServiceStart

avanikp: same here move to Agent

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -139,6 +143,7 @@ public async Task ValidateConnectionStringAsync(string connectionString)
                               "The correct format is: HostName=[hubname];SharedAccessKeyName=" +
                               "[iothubowner or service];SharedAccessKey=[null or valid key]";
                 this.log.Error(message);
+                await this.sendDataToDiagnostics.SendDiagnosticsData("Server Error",message);
```
avanikp: Do we want to log all these messages ? @troyhopwood ?

Services/Diagnostics/SendDataToDiagnostics.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            dynamic jobj = new JObject();
+            jobj.Timestamp = DateTime.Now;
+            jobj.EventType = eventType;
+            if (!string.IsNullOrEmpty(message))
+            {
+                jobj.EventProperties = new JObject(
+                    new JProperty("ErrorMessage", message));
+            }
+            return await httpClient.PostAsync(this.PrepareRequest(ServicesConfig.DIAGNOSTICS_ENDPOINT, jobj));
```
avanikp: pls add error handling

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -11,20 +11,20 @@
 
 namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics
 {
-    public interface ISendDataToDiagnostics
+    public interface IDiagnosticsLogger
     {
-        Task<IHttpResponse> SendDiagnosticsData(string eventType, string message);
+        Task<IHttpResponse> LogDiagnosticsData(string eventType, string message);
     }
```
hathind-ms: add new line

SimulationAgent/SimulationRunner.cs
```diff
@@ -41,6 +41,9 @@ public class SimulationRunner : ISimulationRunner
         // Application logger
         private readonly ILogger log;
 
+        // Logger for logging diagnostics events
+        private readonly IDiagnosticsLogger logDiagnostics;
```
hathind-ms: rename

Services/Diagnostics/SendDataToDiagnostics.cs
```diff
@@ -0,0 +1,60 @@
...
+        private readonly IHttpClient httpClient;
+        
+        public SendDataToDiagnostics(ILogger logger)
+        {
+            this.httpClient = new HttpClient(logger);
+        }
+
+        public async void SendToDiagnostics(object jobj)
+        {
+            var response1 = await httpClient.PostAsync(this.PrepareRequest("http://localhost:9006/v1/diagnosticsevents", jobj));
```
peterfelts: This feels like it should be an environment variable or a constant.

Services/Diagnostics/SendDataToDiagnostics.cs
```diff
@@ -0,0 +1,60 @@
...
+            jobj.DeploymentId = "undefined";
+            jobj.EventType = "Error";
+            jobj.Timestamp = DateTime.Now;
+            this.SendToDiagnostics(jobj);
+        }
+
+        public void SendSimulationDetails(string eventType)
+        {
+            dynamic jobj = new JObject();
+            jobj.DeploymentId = "undefined";
```
peterfelts: What's with these "undefined" deployment ID values?

SimulationAgent/SimulationRunner.cs
peterfelts: Also, is it safe to change the signature of the Start method to async?

Services/CustomDeviceModels.cs
```diff
@@ -42,16 +42,20 @@ public interface ICustomDeviceModels
     public class CustomDeviceModels : ICustomDeviceModels
     {
         private const string STORAGE_COLLECTION = "deviceModels";
+        private const string SERVICE_ERROR = "ServiceError";
```
hathind-ms: Do we have a common Constants file? If so, please move this constant there instead of repeating in every file

Services/CustomDeviceModels.cs
```diff
@@ -67,7 +71,9 @@ public async Task<IEnumerable<DeviceModel>> GetListAsync()
             }
             catch (Exception e)
             {
-                this.log.Error("Unable to load device models from storage", e);
+                string msg = "Unable to load device models from storage";
```
hathind-ms: Why are you creating separate variable for this message when this is used only once? Same applies to other log locations..

Services/Runtime/ConfigData.cs
```diff
@@ -18,13 +18,14 @@ public interface IConfigData
...
         uint? GetOptionalUInt(string key);
+        ILogger GetLogger();
     }
 
     public class ConfigData : IConfigData
     {
         private readonly IConfigurationRoot configuration;
         private readonly ILogger log;
-
+
```
hathind-ms: remove tab space

Services/Runtime/ServicesConfig.cs
```diff
@@ -14,14 +14,17 @@ public interface IServicesConfig
...
+        string DeploymentId { get; }
     }
 
     // TODO: test Windows/Linux folder separator
     //       https://github.com/Azure/device-simulation-dotnet/issues/84
     public class ServicesConfig : IServicesConfig
     {
         public const string USE_DEFAULT_IOTHUB = "default";
-
+
```
hathind-ms: remove tab space

SimulationAgent.Test/SimulationRunnerTest.cs
```diff
@@ -55,12 +57,13 @@ public SimulationRunnerTest(ITestOutputHelper log)
             this.devicePropertiesActor = new Mock<IDevicePropertiesActor>();
             this.rateLimiting = new Mock<IRateLimiting>();
             this.ratingConfig.Setup(x => x.DeviceMessagesPerSecond).Returns(10);
-
+
```
hathind-ms: remove tab space

Services/CustomDeviceModels.cs
sasathy: there is no constants file for services. We do have one for tests but I don't think we might wanna leverage that.

SimulationAgent/SimulationRunner.cs
```diff
@@ -148,7 +161,7 @@ public void Start(Services.Models.Simulation simulation)
             // Use `starting` to exit as soon as possible, to minimize the number 
             // of threads pending on the lock statement
             if (this.starting || this.running) return;
-
+
```
hathind-ms: remove tab

SimulationAgent/SimulationRunner.cs
```diff
@@ -158,7 +171,7 @@ public void Start(Services.Models.Simulation simulation)
                 this.starting = true;
 
                 this.log.Info("Starting simulation...", () => new { simulation.Id });
-
+
```
hathind-ms: remove tab

Services.Test/DiagnosticsLoggerTest.cs
```diff
@@ -0,0 +1,48 @@
...
+            this.mockHttpClient
+                .Setup(x => x.PostAsync(It.IsAny<IHttpRequest>()))
+                .ReturnsAsync(response);
+
+            // Act
+            IHttpResponse result = diagnosticsLogger.LogDiagnosticsData("ServiceError", "").Result;
+
+            // Assert - Testing to see if the logic in the function is working fine. 
+            // So, asserting if the expected response and actual responses are similar.
+            Assert.Equal(response,result);
```
avanikp: nit: add space after ,

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+            dynamic jobj = new JObject();
+            jobj.EventId = Guid.NewGuid().ToString();
+            jobj.EventType = eventType;
+            jobj.DeploymentId = this.servicesConfig.DeploymentId;
+            jobj.SolutionType = this.servicesConfig.SolutionType;
+            if (!string.IsNullOrEmpty(message))
+            {
+                jobj.EventProperties = new JObject(
+                    new JProperty("ErrorMessage", message));
+            }
```
avanikp: nit: add nre line after }

Services/DeviceMethods.cs
```diff
@@ -26,20 +26,25 @@ public class DeviceMethods : IDeviceMethods
...
         private readonly ILogger log;
+        private readonly IDiagnosticsLogger diagnosticsLogger;
+        
         private readonly IScriptInterpreter scriptInterpreter;
         private IDictionary<string, Script> cloudToDeviceMethods;
         private ISmartDictionary deviceState;
         private ISmartDictionary deviceProperties;
         private string deviceId;
         private bool isRegistered;
+        private const string SERVICE_ERROR = "ServiceError";
```
avanikp: nit: suggest renaming this to SERVICE_ERROR_EVENT

Services/Runtime/ConfigData.cs
```diff
@@ -36,6 +37,10 @@ public string GetString(string key, string defaultValue = "")
             return this.GetStringInternal(key, defaultValue);
         }
 
+        public ILogger GetLogger()
+        {
+            return this.log;
+        }
```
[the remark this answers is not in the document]
sasathy: done

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+            return await httpClient.PostAsync(this.PrepareRequest(this.servicesConfig.DiagnosticsEndpointUrl, jobj));
+        }
+
+        private HttpRequest PrepareRequest(string path, object obj=null)
+        {
+            var request = new HttpRequest();
+            request.AddHeader(HttpRequestHeader.Accept.ToString(), "application/json");
```
avanikp: Is this required ?

Services/Diagnostics/DiagnosticsLogger.cs
sasathy: done

Services/CustomDeviceModels.cs
```diff
@@ -42,16 +42,20 @@ public interface ICustomDeviceModels
     public class CustomDeviceModels : ICustomDeviceModels
     {
         private const string STORAGE_COLLECTION = "deviceModels";
+        private const string SERVICE_ERROR_EVENT = "ServiceError";
```
dluc: I see the same constant is present in multiple files, it will make the code easier to maintain if we group all the diagnostics constants in one place. For instance I would create a `DiagnosticsConstants.cs` file in the `Diagnostics` folder with this any other value. e.g.

```
public class DiagnosticsConstants
{
    public const string SERVICE_ERROR_EVENT = "ServiceError";
}

File: Services/CustomDeviceModels.cs [hunk 29]
```diff
@@ -67,7 +71,9 @@ public async Task<IEnumerable<DeviceModel>> GetListAsync()
             }
             catch (Exception e)
             {
-                this.log.Error("Unable to load device models from storage", e);
+                var msg = "Unable to load device models from storage";
+                this.log.Error(msg, e);
+                this.diagnosticsLogger.LogDiagnosticsData(SERVICE_ERROR_EVENT, $"{msg}: {e.Message}");
```
dluc: I would suggest using a more semantic approach, e.g. instead of a generic `LogDiagnosticsData(..)` method, have a `LogServiceError(..)` method in the logger. This way you don't need to know which constant to use and the code is more readable. I would also delegate to the logger to extract information from the exception, passing the full object:

```
this.diagnosticsLogger.LogServiceError(msg, e)
```

Services/CustomDeviceModels.cs
dluc: Anyhow, as per other comment about using a semantic log approach, I would actually move this constant (and keep it private) inside the logger class.

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+        private readonly IHttpClient httpClient;
+        private readonly IServicesConfig servicesConfig;
+              
+        public DiagnosticsLogger(IHttpClient httpClient, IServicesConfig servicesConfig)
+        {
+            this.httpClient = httpClient;
+            this.servicesConfig = servicesConfig;
+        }
+
+        public async Task<IHttpResponse> LogDiagnosticsData(string eventType, string message = "")
```
dluc: please add the same `Caller...` params used in `Logger.cs` to capture information about the filename where the event is coming from. Hopefully the message is unique enough, but I would include `callerName`, `filePath` and `lineNumber` at the end of the msg, e.g. in brackets

dluc: async methods should have the`Async` suffix and are usually awaited. Is it intentional not to?

Services/Devices.cs
```diff
@@ -298,17 +308,23 @@ public async Task DeleteListAsync(IEnumerable<string> deviceIds)
...
+                this.log.Error(msg, () => new { error.InnerException, error });
+                this.diagnosticsLogger.LogDiagnosticsData(SERVICE_ERROR_EVENT, $"{msg}: {error.Message}");
                 throw;
             }
             catch (Exception error)
             {
-                this.log.Error("Failed to delete devices", error);
+                var msg = "Failed to delete devices";
+                this.log.Error(msg, error);
+                this.diagnosticsLogger.LogDiagnosticsData(SERVICE_ERROR_EVENT, $"{msg}: {error.Message}"); ;
```
dluc: this is an async method, is it intentional not to wait it?

I know it's a very different, but I would have used a different approach:
1. when an error occurs, give it to the logger
2. the logger keeps the events in memory, it doesn't send them immediately
3. the logger has a thread that every 2-3 seconds check if there are events to send, and send them

Do you think this could be done in future?

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            dynamic jobj = new JObject();
```
dluc: please avoid `dynamic` as a rule of thumb - it's a very slow and expensive type affecting the service performance.  You can define a `struct` to hold all the fields below

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        private HttpRequest PrepareRequest(string path, object obj=null)
+        {
+            var request = new HttpRequest();
+            request.SetUriFromString(path);
+            request.SetContent(obj);
+            return request;
+        }
+
```
dluc: remove empty line

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+            if (!string.IsNullOrEmpty(message))
+            {
+                jobj.EventProperties = new JObject(
+                    new JProperty("ErrorMessage", message));
+            }
+
+            return await httpClient.PostAsync(this.PrepareRequest(this.servicesConfig.DiagnosticsEndpointUrl, jobj));
+        }
+
+        private HttpRequest PrepareRequest(string path, object obj=null)
```
dluc: spacing, check code style

Services.Test/DiagnosticsLoggerTest.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Http;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Runtime;
+using Moq;
+using Services.Test.helpers;
+using Xunit;
+
+namespace Services.Test
```
dluc: the folder structure of test classes should match that of the classes under test. This convention makes it easy to find code and tests, and also allows to organize tests logically (as opposed to have a long list of files in the same folder).
In this case, there should be a `Diagnostics` folder, containing this class, and the namespace should be `Services.Test.Diagnostics`.

Please check also the other files if they are in the right folder

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,160 @@
...
+    {
+        public struct JsonStruct
+        {
+            public string EventId;
+            public string EventType;
+            public string DeploymentId;
+            public string SolutionType;
+            public Dictionary<string, object> EventProperties;
+
+            public JsonStruct(string st1, string st2, string st3, string st4, Dictionary<string, object> eventProps)
```
hathind-ms: remove 1,2,3,4 from variable names. Give them more meaningful names

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,160 @@
...
+            object data,
+            [CallerMemberName] string callerName = "",
+            [CallerFilePath] string filePath = "",
+            [CallerLineNumber] int lineNumber = 0)
+        {
+            Dictionary<string, object> eventProps = new Dictionary<string, object>();
+            eventProps.Add("message", message + $"(CallerMemberName = '{callerName}', CallerFilePath = '{filePath}', lineNumber = '{lineNumber}')");
+            if (e != null)
+            {
+                eventProps.Add("Exception", e);
```
hathind-ms: do we need to send full exception to Diagnostics or just Exception message is fine? For Diagnostics - exception message alone should be enough alongwith filePath and lineNumber
@dluc/ @troyhopwood  - Can you confirm?

Services/CustomDeviceModels.cs
```diff
@@ -67,7 +70,9 @@ public async Task<IEnumerable<DeviceModel>> GetListAsync()
             }
             catch (Exception e)
             {
-                this.log.Error("Unable to load device models from storage", e);
+                var msg = "Unable to load device models from storage";
+                this.log.Error(msg, e);
+                this.diagnosticsLogger.LogServiceErrorAsync(msg, e);
                 throw new ExternalDependencyException("Unable to load device models from storage", e);
```
hathind-ms: use msg instead

Services/CustomDeviceModels.cs
```diff
@@ -86,7 +91,9 @@ public async Task<IEnumerable<DeviceModel>> GetListAsync()
             }
             catch (Exception e)
             {
-                this.log.Error("Unable to parse device models loaded from storage", e);
+                var msg = "Unable to parse device models loaded from storage";
+                this.log.Error(msg, e);
+                this.diagnosticsLogger.LogServiceErrorAsync(msg, e);
                 throw new ExternalDependencyException("Unable to parse device models loaded from storage", e);
```
hathind-ms: use msg instead. Same applies below where you are logging and throwing exception

Services/Simulations.cs
```diff
@@ -188,7 +191,9 @@ public async Task<Models.Simulation> UpsertAsync(Models.Simulation simulation)
 
                 if (simulation.ETag != simulations[0].ETag)
                 {
-                    this.log.Error("Invalid ETag. Running simulation ETag is:'", () => new { simulations });
+                    var msg = "Invalid ETag. Running simulation ETag is:'";
+                    this.log.Error(msg, () => new { simulations });
+                    this.diagnosticsLogger.LogServiceErrorAsync(msg, new { simulations });
                     throw new ResourceOutOfDateException("Invalid ETag. Running simulation ETag is:'" + simulations[0].ETag + "'.");
```
hathind-ms: use msg instead

Services/StockDeviceModels.cs
```diff
@@ -68,7 +72,9 @@ public IEnumerable<DeviceModel> GetList()
             }
             catch (Exception e)
             {
-                this.log.Error("Unable to parse Device Model files", e);
+                var msg = "Unable to parse Device Model files";
+                this.log.Error(msg, e);
+                this.diagnosticsLogger.LogServiceErrorAsync(msg, e);
                 throw new InvalidConfigurationException("Unable to parse Device Model files: " + e.Message, e);
```
hathind-ms: use msg instead

SimulationAgent/Agent.cs
```diff
@@ -127,6 +142,7 @@ private void CheckForNewSimulation(Services.Models.Simulation newSimulation)
                 if (this.simulation.ShouldBeRunning())
                 {
                     this.log.Debug("------ Starting new simulation ------", () => new { this.simulation });
+                    this.logDiagnostics.LogServiceStartAsync();
                     this.runner.Start(this.simulation);
                     this.log.Debug("------ New simulation started ------", () => new { this.simulation });
```
hathind-ms: Log Simulation Start event too below

SimulationAgent/SimulationRunner.cs
```diff
@@ -591,8 +607,10 @@ private void TryToStartTelemetryThreads()
             }
             catch (Exception e)
             {
+                var msg = "Unable to start the telemetry threads";
                 this.IncrementSimulationErrorsCount();
-                this.log.Error("Unable to start the telemetry threads", e);
+                this.log.Error(msg, e);
+                this.diagnosticsLogger.LogServiceErrorAsync(msg, e);
                 throw new Exception("Unable to start the telemetry threads", e);
```
hathind-ms: use msg instead. Check other places/files too

Services/Diagnostics/DiagnosticsLogger.cs
sasathy: I see for myself that I am naming variables in a random manner. Will take extra care from next time. Sorry about this.

Services.Test/Diagnostics/DiagnosticsLoggerTest.cs
```diff
@@ -0,0 +1,106 @@
...
+            IHttpResponse logging_service_error_message_and_exception = diagnosticsLogger.LogServiceErrorAsync("testmessage", new System.Exception()).Result;
+            // Logging service error along with an object
+            IHttpResponse logging_service_error_message_and_object = diagnosticsLogger.LogServiceErrorAsync("testmessage", new { Test = "test" }).Result;
+
+            // Assert - Testing to see if the logic in the function is working fine. 
+            // So, asserting if the expected response and actual responses are similar.
+            Assert.Equal(response, logging_service_error_message);
+            Assert.Equal(response, logging_service_error_message_and_exception);
+            Assert.Equal(response, logging_service_error_message_and_object);
+            //Assert.Equal(response, result3);
```
hathind-ms: remove this commented line

WebService/appsettings.ini
```diff
@@ -124,6 +124,9 @@ ExtraDiagnostics = false
 # be created or is not writable.
 ExtraDiagnosticsPath = "/tmp/sim/"
 
+# Path to the diagnostics back end where all telemetry data is to be directed
+diagnostic_endpoint_url = "${?PCS_DIAGNOSTICS_WEBSERVICE_URL}"
```
dluc: "diagnostics"

WebService/v1/Controllers/StatusController.cs
```diff
@@ -109,8 +113,10 @@ public async Task<StatusApiModel> GetAsync()
             }
             catch (Exception e)
             {
-                errors.Add("Unable to fetch simulation status");
-                this.log.Error("Unable to fetch simulation status", e);
+                var msg = "Unable to fetch simulation status";
```
dluc: we're removing information about the simulation from this endpoint, so we might not need this. Could you double check with @hathind-ms ?

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,146 @@
...
+            [CallerLineNumber] int lineNumber = 0);
+
+        Task<IHttpResponse> LogServiceErrorAsync(
+            string message,
+            object data,
+            [CallerMemberName] string callerName = "",
+            [CallerFilePath] string filePath = "",
+            [CallerLineNumber] int lineNumber = 0);
+
+        Task<IHttpResponse> LogServiceExceptionAsync(
```
dluc: what's the difference between and Error and and Exception?

dluc: I think we could merge the 2, or at least use the same name. When logging an exception we might want to attache data, e.g. exceptionMessage could be part of data

WebService/Runtime/Config.cs
```diff
@@ -109,14 +111,17 @@ public class Config : IConfig
...
+
         public int Port { get; }
         public ILoggingConfig LoggingConfig { get; set; }
         public IClientAuthConfig ClientAuthConfig { get; }
         public IServicesConfig ServicesConfig { get; }
         public IRateLimitingConfig RateLimitingConfig { get; set; }
         public IDeploymentConfig DeploymentConfig { get; set; }
         public IConcurrencyConfig ConcurrencyConfig { get; set; }
-
+        public IDiagnosticsLogger DiagnosticsLogger { get; set; }
```
dluc: the `Config` class is not supposed to provide dependencies like the diagnostics logger, I think we need to remove this and the http client and the `GetLogger` method from configData. Not sure why you chose this approach, but dependencies should be provided via the DI.

WebService/v1/Controllers/StatusController.cs
hathind-ms: don't need to log here.

WebService/appsettings.ini
sasathy: done

WebService/v1/Controllers/StatusController.cs
sasathy: done

Services/Diagnostics/DiagnosticsLogger.cs
sasathy: renamed it :) done

WebService/Runtime/Config.cs
sasathy: done.

WebService/v1/Controllers/StatusController.cs
```diff
@@ -130,8 +132,9 @@ public async Task<StatusApiModel> GetAsync()
             }
             catch (Exception e)
             {
+                var msg = "Storage ping failed";
```
maple10001: nit: revert changes, if not using the variable `msg` multiple times.

sasathy:
Closing this PR since git is unable to receive the status from Travis

SimulationAgent/SimulationRunner.cs
```diff
@@ -38,9 +40,15 @@ public class SimulationRunner : ISimulationRunner
         // Allow 30 seconds to create the devices (1000 devices normally takes 2-3 seconds)
         private const int DEVICES_CREATION_TIMEOUT_SECS = 30;
 
+        // Sending service heartbeat to diagnostics backend every 24 hours/ 1 day
+        private const int DIAGNOSTICS_POLLING_FREQUENCY = 1;
```
dluc: not used. should it be? the comment doesn't match the variable name...

SimulationAgent/Agent.cs
```diff
@@ -24,31 +24,37 @@ public interface ISimulationAgent
     public class Agent : ISimulationAgent
     {
         private const int CHECK_INTERVAL_MSECS = 10000;
+        private const int DIAGNOSTICS_POLLING_FREQUENCY = 1;
```
dluc: pls add a suffix to explain if the value is in seconds, hours, etc.

SimulationAgent/Agent.cs
```diff
@@ -59,6 +65,16 @@ public async Task RunAsync()
             while (this.running)
             {
                 var oldSimulation = this.simulation;
+                DateTime now = DateTime.Now;
```
dluc: use `DateTimeOffset` and `UtcNow`

SimulationAgent/Agent.cs
```diff
@@ -59,6 +65,16 @@ public async Task RunAsync()
...
+                TimeSpan duration = now - this.lastPolledTime;
+                
+                // Send heartbeat every 24 hours
+                if (duration.Days >= DIAGNOSTICS_POLLING_FREQUENCY)
+                {
+                    this.lastPolledTime = DateTime.Now;
```
dluc: ditto

SimulationAgent/Agent.cs
```diff
@@ -59,6 +65,16 @@ public async Task RunAsync()
...
+                    this.logDiagnostics.LogServiceHeartbeatAsync();
+                }
```
dluc: pls move all the heartbeat code into a separate private method, and call it from here. e.g. `await this.SendSolutionHeartbeatAsync()`

you could also consider moving the logic into the DiagnosticsLogger instead of having it here

SimulationAgent/Agent.cs
```diff
@@ -205,9 +221,13 @@ private void CheckForNewSimulation(Simulation newSimulation)
                 this.simulation = newSimulation;
                 if (this.simulation.ShouldBeRunning())
                 {
-                    this.log.Debug("------ Starting new simulation ------", () => new { this.simulation });
+                    var simulationStatus = "------ Starting new simulation ------";
```
dluc: this message is styled with "------" to be visible in the logs, but I doubt that's what we want to see in the backend

SimulationAgent/Agent.cs
```diff
@@ -205,9 +221,13 @@ private void CheckForNewSimulation(Simulation newSimulation)
...
+                    this.log.Debug(simulationStatus, () => new { this.simulation });
+                    this.logDiagnostics.LogServiceStartAsync(simulationStatus);
                     this.runner.Start(this.simulation);
-                    this.log.Debug("------ New simulation started ------", () => new { this.simulation });
+                    simulationStatus = "------ New simulation started ------";
```
dluc: ditto

SimulationAgent/Agent.cs
```diff
@@ -59,6 +65,16 @@ public async Task RunAsync()
...
             {
                 var oldSimulation = this.simulation;
+                DateTime now = DateTime.Now;
+                TimeSpan duration = now - this.lastPolledTime;
+                
+                // Send heartbeat every 24 hours
+                if (duration.Days >= DIAGNOSTICS_POLLING_FREQUENCY)
+                {
+                    this.lastPolledTime = DateTime.Now;
+                    this.logDiagnostics.LogServiceHeartbeatAsync();
```
dluc: pls add test for this code. We should verify that the heartbeat is sent only once a day, and not for every loop (which would DDOS the endpoint)

Services/Simulations.cs
```diff
@@ -209,8 +212,10 @@ public async Task<Models.Simulation> UpsertAsync(Models.Simulation simulation)
 
                 if (simulation.ETag != existingSimulation.ETag)
                 {
-                    this.log.Error("Invalid ETag. Running simulation ETag is:'", () => new { simulation });
-                    throw new ResourceOutOfDateException("Invalid ETag. Running simulation ETag is:'" + simulation.ETag + "'.");
+                    var msg = "Invalid ETag. Running simulation ETag is:'";
+                    this.log.Error(msg, () => new { simulation });
+                    this.diagnosticsLogger.LogServiceErrorAsync(msg, new { simulation });
```
dluc: logging the entire simulation object risks sending sensitive data. Please double check this with @hathind-ms and see if we can limit the data sent. We should check the entire codebase for the same, making sure this is not sending sensitive and unneeded data

Services.Test/Diagnostics/DiagnosticsLoggerTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            this.mockHttpClient
+                .Setup(x => x.PostAsync(It.IsAny<IHttpRequest>()))
+                .ReturnsAsync(response);
+
+            // Act
+            IHttpResponse result = diagnosticsLogger.LogServiceStartAsync("test").Result;
+
+            // Assert - Testing to see if the logic in the function is working fine. 
+            // So, asserting if the expected response and actual responses are similar.
+            Assert.Equal(response, result);
```
dluc: `response` is an empty object (line 27). What is the equality testing?

Services.Test/Diagnostics/DiagnosticsLoggerTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            this.mockHttpClient
+                .Setup(x => x.PostAsync(It.IsAny<IHttpRequest>()))
+                .ReturnsAsync(response);
+
+            // Act
+            IHttpResponse result = diagnosticsLogger.LogServiceHeartbeatAsync().Result;
+
+            // Assert - Testing to see if the logic in the function is working fine. 
+            // So, asserting if the expected response and actual responses are similar.
+            Assert.Equal(response, result);
```
dluc: same here...

Services.Test/Diagnostics/DiagnosticsLoggerTest.cs
```diff
@@ -0,0 +1,105 @@
...
+                    DiagnosticsEndpointUrl = DIAGNOSTICS_SERVICE_URL
+                });
+
+            this.mockHttpClient
+                .Setup(x => x.PostAsync(It.IsAny<IHttpRequest>()))
+                .ReturnsAsync(response);
+
+            // Act
+            // Logging service error sending just a message string
+            IHttpResponse logging_service_error_message = diagnosticsLogger.LogServiceErrorAsync("testmessage").Result;
```
dluc: code style: we don't use snake_case, please use camelCase as for the other variables

Services.Test/Diagnostics/DiagnosticsLoggerTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            // Logging service error along with an exception
+            IHttpResponse logging_service_error_message_and_exception = diagnosticsLogger.LogServiceErrorAsync("testmessage", new System.Exception().Message).Result;
+            // Logging service error along with an object
+            IHttpResponse logging_service_error_message_and_object = diagnosticsLogger.LogServiceErrorAsync("testmessage", new { Test = "test" }).Result;
+
+            // Assert - Testing to see if the logic in the function is working fine. 
+            // So, asserting if the expected response and actual responses are similar.
+            Assert.Equal(response, logging_service_error_message);
+            Assert.Equal(response, logging_service_error_message_and_exception);
+            Assert.Equal(response, logging_service_error_message_and_object);
```
dluc: I'm not sure what these assertions are actually testing

Services/Services.csproj
```diff
@@ -10,6 +10,7 @@
     <PackageReference Include="jint" Version="2.10.4" />
     <PackageReference Include="Microsoft.Azure.Devices" Version="1.16.1" />
     <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.17.1" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
```
dluc: could you verify what this is needed for? if possible, remove this reference

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Http;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Runtime;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics
+{
+    public interface IDiagnosticsLogger
+    {
+        Task<IHttpResponse> LogServiceStartAsync(string message);
```
dluc: all these methods return `IHttpResponse`. Is the response actually used anywhere, other than for testing?  If it's just for testing, we should remove it. In fact, a class calling a logger should not know how the logs are collected (in memory, file, network etc.).  Probably this class will soon collect events in memory and send them out in batch, so there won't be a distinct and immediate http response for every event logged.

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        private HttpRequest PrepareRequest(string path, object obj = null)
+        {
+            var request = new HttpRequest();
+            request.SetUriFromString(path);
+            request.SetContent(obj);
+            return request;
+        }
+    }
+}
```
dluc: check code style (spacing etc.) in the file

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        private const string SERVICE_ERROR_EVENT = "ServiceError";
+        private const string SERVICE_START_EVENT = "ServiceStart";
+        private const string SERVICE_HEARTBEAT_EVENT = "ServiceHeartbeat";
+        
+        public DiagnosticsLogger(IHttpClient httpClient, IServicesConfig servicesConfig)
+        {
+            this.httpClient = httpClient;
+            this.servicesConfig = servicesConfig;
+            this.diagnosticsEndpoint = this.servicesConfig.DiagnosticsEndpointUrl + "/diagnosticsevents";
```
dluc: can we move this to the configuration file? e.g. `diagnostics_endpoint_url = "${?PCS_DIAGNOSTICS_WEBSERVICE_URL}/diagnosticsevents"`.  I rather not hardcode this, so it's easier to swap the service out with another one that used a different URL

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,128 @@
...
+            [CallerLineNumber] int lineNumber = 0)
+        {
+            JsonStruct jsonStruct = this.ConvertToJson(message, "", data, callerName, filePath, lineNumber);
+            return await this.httpClient.PostAsync(this.PrepareRequest(this.diagnosticsEndpoint, jsonStruct));
+        }
+
+        private JsonStruct ConvertToJson(string message,
+            string exceptionMessage,
+            object data,
+            [CallerMemberName] string callerName = "",
```
dluc: this method should not use the [Caller...] attributes

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,128 @@
...
+            [CallerFilePath] string filePath = "",
+            [CallerLineNumber] int lineNumber = 0)
+        {
+            JsonStruct jsonStruct = this.ConvertToJson(message, "", data, callerName, filePath, lineNumber);
+            return await this.httpClient.PostAsync(this.PrepareRequest(this.diagnosticsEndpoint, jsonStruct));
+        }
+
+        private JsonStruct ConvertToJson(string message,
```
dluc: nit: rename to `ConvertServiceErrorToJson` (because of line 117, this method cannot be reused to serialize other types of event)

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -0,0 +1,128 @@
...
+            [CallerFilePath] string filePath = "",
+            [CallerLineNumber] int lineNumber = 0);        
+    }
+
+    public class DiagnosticsLogger : IDiagnosticsLogger
+    {
+        public struct JsonStruct
+        {
+            public string EventType;
+            public Dictionary<string, object> EventProperties;
```
dluc: to make sure these two properties are serialized correctly when sent out, pls add `[JsonProperty(PropertyName = "...")]` on each

SimulationAgent/Agent.cs
```diff
@@ -198,6 +207,19 @@ private void CheckForChangedSimulation(Services.Models.Simulation newSimulation)
...
 
+        private void SendSolutionHeartbeatAsync()
+        {
+            DateTimeOffset now = DateTimeOffset.UtcNow;
+            TimeSpan duration = now - this.lastPolledTime;
+
+            // Send heartbeat every 24 hours
+            if (duration.Days >= 1)
+            {
+                this.lastPolledTime = DateTimeOffset.UtcNow;
```
dluc: you could reuse the var created at line 212

SimulationAgent/Agent.cs
```diff
@@ -24,31 +24,37 @@ public interface ISimulationAgent
     public class Agent : ISimulationAgent
     {
         private const int CHECK_INTERVAL_MSECS = 10000;
+        private const int DIAGNOSTICS_POLLING_FREQUENCY_DAYS = 1;
```
dluc: where is this used? I assume it should be used at line 216 right?

SimulationAgent/Agent.cs
```diff
@@ -206,8 +228,10 @@ private void CheckForNewSimulation(Simulation newSimulation)
                 if (this.simulation.ShouldBeRunning())
                 {
                     this.log.Debug("------ Starting new simulation ------", () => new { this.simulation });
+                    this.logDiagnostics.LogServiceStartAsync("Starting new simulation");
```
hathind-ms: add Simulation Id to the message. Here and in line 234

PR #247, opened 2018-09-10
Bug fixes

hathind-ms:
# Description and Motivation 
- Setting default IoTHubConnection string
- Put should not throw exception if there is no existing simulation with the requested id.
...

# Change type 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

**Checklist:**

- [ ] All tests passed

### DIFF
--- a/Services/Models/Simulation.cs
+++ b/Services/Models/Simulation.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models
         public IList<string> IotHubConnectionStrings
         {
             get => this.iotHubConnectionStrings;
-            set => this.iotHubConnectionStrings = value ?? new List<string> { };
+            set => this.iotHubConnectionStrings = value ?? new List<string> { ServicesConfig.USE_DEFAULT_IOTHUB };
         }
 
         public Simulation()
@@ -62,7 +62,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models
             this.EndTime = DateTimeOffset.MaxValue;
 
             // by default, use environment variable
-            this.IotHubConnectionStrings = new List<string> { };
+            this.IotHubConnectionStrings = new List<string> { ServicesConfig.USE_DEFAULT_IOTHUB };
             this.DeviceModels = new List<DeviceModelRef>();
             this.CustomDevices = new List<CustomDeviceRef>();
             this.Statistics = new StatisticsRef();
--- a/Services/Simulations.cs
+++ b/Services/Simulations.cs
@@ -196,7 +196,18 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
                 throw new InvalidInputException("Simulation ID is not specified.");
             }
 
-            var existingSimulation = await this.GetAsync(simulation.Id);
+            Models.Simulation existingSimulation = null;
+
+            try
+            {
+                existingSimulation = await this.GetAsync(simulation.Id);
+            }
+            catch (ResourceNotFoundException)
+            {
+                // Do nothing - this mean simulation does not exist in storage
+                // Therefore allow to create a new simulation
+            }
+
             if (existingSimulation != null)
             {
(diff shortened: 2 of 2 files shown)


Services/Simulations.cs
```diff
@@ -199,7 +199,18 @@ public async Task<Models.Simulation> UpsertAsync(Models.Simulation simulation)
...
-            var existingSimulation = await this.GetAsync(simulation.Id);
+            Models.Simulation existingSimulation = null;
+
+            try
+            {
+                existingSimulation = await this.GetAsync(simulation.Id);
+            }
+            catch (ResourceNotFoundException)
+            {
+                // Do nothing - this mean simulation does not exist in storage
```
avanikp: suggest adding log.Info here

Services/Diagnostics/DiagnosticsLogger.cs
```diff
@@ -131,10 +124,29 @@ public void LogServiceHeartbeat()
...
+            try
+            {
+                var request = new HttpRequest();
+                request.SetUriFromString(path);
+                request.SetContent(obj);
+                return request;
+            }
+            catch
+            {
+                // Failed to construct uri
```
avanikp: same here.. log.info the exception

Services/Models/Simulation.cs
```diff
@@ -57,7 +57,7 @@ public class Simulation
         public IList<string> IotHubConnectionStrings
         {
             get => this.iotHubConnectionStrings;
-            set => this.iotHubConnectionStrings = value ?? new List<string>();
+            set => this.iotHubConnectionStrings = value ?? new List<string> { ServicesConfig.USE_DEFAULT_IOTHUB };
```
peterfelts: Why is this needed here? I see that there is code that appears to be setting this in the API model, perhaps we should set the default there if we're getting to a point where the default isn't set properly in the service model.

Services/Simulations.cs
```diff
@@ -192,7 +192,19 @@ public async Task<Models.Simulation> UpsertAsync(Models.Simulation simulation)
                 throw new InvalidInputException("Simulation ID is not specified.");
             }
 
-            var existingSimulation = await this.GetAsync(simulation.Id);
+            Models.Simulation existingSimulation = null;
```
peterfelts: If the simulation ID doesn't exist, the caller should be calling InsertAsync, not UpsertAsync. Perhaps we should leave this unchanged and determine why UpsertAsync is being called on a simulation that doesn't exist.

peterfelts: We shouldn't get in to a state where UpsertAsync is being called for a simulation that doesn't exist. InsertAsync should be called. Under what scenario are we seeing that UpsertAsync is being called for a simulation that doesn't exist? I could see that possibly we'd get in to a race condition where a simulation was deleted right before being upserted, but in that case we should not re-create the simulation.

Services/Models/Simulation.cs
hathind-ms: Setting default value here. We used to have a default value which changed with my changes

Services/Simulations.cs
hathind-ms: resolved offline

Services/Simulations.cs
```diff
@@ -167,12 +167,14 @@ public async Task<Models.Simulation> InsertAsync(Models.Simulation simulation, s
...
+            for (var index = 0; index < simulation.IotHubConnectionStrings.Count; index++)
             {
-                var connString = await this.connectionStringManager.RedactAndStoreAsync(iotHubConnectionString);
-                simulation.IotHubConnectionStrings.Add(connString);
+                var connString = await this.connectionStringManager.RedactAndStoreAsync(simulation.IotHubConnectionStrings[index]);
+
+                if (!simulation.IotHubConnectionStrings.Contains(connString))
+                {
+                    simulation.IotHubConnectionStrings[index] = connString;
+                }
```
dluc: aside from changing from `foreach` to `for` the change here is just to avoid duplicate conn string, right?

PR #251, opened 2018-09-12
Add agent for creating and partitioning devices

dluc:
# Description and Motivation 

Add a new agent to the solution, which will be responsible for creating the devices in IoT hub and for partitioning the list of devices, so that multiple VMs can each simulate a different set of devices.
The agent is not started yet because the new storage library is not ready.

The PR includes some config refactoring and cleanup, e.g. having all the storage settings in the same place, and moving the entire diagnostics url to config (instead of being partially hard coded) 

# Change type 

- [ ] Bug fix
- [x] New feature
- [x] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/251)

### DIFF
--- a/PartitioningAgent/Agent.cs
+++ b/PartitioningAgent/Agent.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading.Tasks;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Clustering;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Concurrency;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.PartitioningAgent
+{
+    public interface IPartitioningAgent
+    {
+        Task StartAsync();
+        void Stop();
+    }
+
+    public class Agent : IPartitioningAgent
+    {
+        private readonly IClusterNodes clusterNodes;
+        private readonly IThreadWrapper thread;
+        private readonly ILogger log;
+        private readonly int checkIntervalMsecs;
+
+        private bool running;
+
+        public Agent(
+            IClusterNodes cluster,
+            IThreadWrapper thread,
+            IClusteringConfig clusteringConfig,
+            ILogger logger)
+        {
+            this.clusterNodes = cluster;
+            this.thread = thread;
+            this.log = logger;
+            this.checkIntervalMsecs = clusteringConfig.CheckIntervalMsecs;
+            this.running = false;
+        }
+
+        public async Task StartAsync()
+        {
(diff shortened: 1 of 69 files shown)


WebService/Startup.cs
```diff
@@ -18,6 +19,10 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService
 {
     public class Startup
     {
+        // Agent responsible for creating devices and partitions
+        private IPartitioningAgent partitioningAgent;
+
+        // Agent responsible for simulating IoT devicesw
```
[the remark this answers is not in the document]
dluc: fixed

WebService/Startup.cs
```diff
@@ -77,28 +82,34 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
+            this.partitioningAgent = this.ApplicationContainer.Resolve<IPartitioningAgent>();
+
+            // TODO: uncomment when ready
+            //this.partitioningAgent.StartAsync();
+
             this.simulationAgent = this.ApplicationContainer.Resolve<ISimulationAgent>();
             this.simulationAgent.RunAsync();
         }
 
         private void StopAgent()
```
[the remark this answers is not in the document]
dluc: done

PartitioningAgent/Agent.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        private bool running;
+
+        public Agent(
+            IClusterNodes cluster,
+            IThreadWrapper thread,
+            IClusteringConfig clusteringConfig,
+            ILogger logger)
+        {
+            this.clusterNodes = cluster;
```
peterfelts: Where does this get passed in from? If Autofac is doing it, I'm not clear on where the values are coming from.

Services/Clustering/ClusterNodes.cs
```diff
@@ -0,0 +1,45 @@
...
+        public Task KeepAliveNodeAsync()
+        {
+            throw new NotImplementedException("Need the new storage code first");
+        }
+
+        // Generate a unique value used to identify the current instance
+        private static string GenerateSharedNodeId()
+        {
+            // Example: 12a34b5678901c23de4c5f6ab78c9012.2018-01-12T01:15:00
+            return Guid.NewGuid().ToString("N") + "." + DateTimeOffset.UtcNow.ToString("s");
```
peterfelts: What's the reason for having the date added to the node ID? Is it just for readability?

PartitioningAgent/Agent.cs
dluc: I don't understand the question, `cluster` is injected as usual, see line 26

Services/Clustering/ClusterNodes.cs
dluc: only for debugging purpose. When looking at logs containing node Id, each node will have a random id plus a datetime indicating when the node started

PR #252, opened 2018-09-13
Maintenance and clean up some build warnings

dluc:
# Description and Motivation 

* Maintenance: upgrade some dependencies
* Fix some compilation warnings due syntax used with non-blocking code

# Change type 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

**Checklist:**

- [x] All tests passed

### DIFF
--- a/Services/Simulations.cs
+++ b/Services/Simulations.cs
@@ -69,19 +69,22 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         private readonly IIotHubConnectionStringManager connectionStringManager;
         private readonly IDevices devices;
         private readonly ILogger log;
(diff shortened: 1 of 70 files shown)


Services/Simulations.cs
```diff
@@ -214,7 +214,7 @@ public async Task<Models.Simulation> UpsertAsync(Models.Simulation simulation)
                 {
                     var msg = "Invalid ETag. Running simulation ETag is:'";
```
hathind-ms: Not related to your change but we should fix it. Append ETag to the message, otherwise log message is incomplete

dluc: 👍  fixed

PR #253, opened 2018-09-13
Manage device simulation nodes

dluc:
# Description and Motivation 

When a node starts it generates a unique ID to identify itself, and writes the ID into the Nodes collection. The ID expires unless the process keeps updating the record. The expiration happens when loading the list of nodes. As a result, the nodes collection contains an eventually consistent list of active nodes, which can collaborate to run simulations.
To avoid contentions, only one node takes care of cleaning up the nodes collection. In order to decide which node, there is a master-node election process. A node tries to become master locking a record, and holding the lock as long as it is alive. If the master node crashes, the lock expires, and another node eventually becomes master. The master election process will be used also to decide which node creates devices and partitions.

# Change type 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/253)

### DIFF
--- a/WebService/appsettings.ini
+++ b/WebService/appsettings.ini
@@ -171,38 +171,50 @@ twin_writes_per_second = 10
 # A value too high (e.g. > 10) can decrease the overall throughput due to context switching.
 # A value too low (e.g. < 2) can decrease the overall throughput due to the time required to
 # loop through all the devices, when dealing we several thousands of devices.
+# The default value is 4
 telemetry_threads = 4
 
 # While connecting all the devices, limit the number of connections waiting to be
 # established. A low number will slow down the time required to connect all the devices.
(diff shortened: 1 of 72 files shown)


WebService/appsettings.ini
```diff
@@ -202,45 +202,45 @@ twin_writes_per_second = 10
 # A value too high (e.g. > 10) can decrease the overall throughput due to context switching.
 # A value too low (e.g. < 2) can decrease the overall throughput due to the time required to
 # loop through all the devices, when dealing we several thousands of devices.
```
hathind-ms: typo: 'with' instead of 'we'

dluc: fixed

PR #255, opened 2018-09-17
Create devices bulk import job

dluc:
# Description and Motivation

Create devices in a more scalable way using Azure IoT Hub device creation jobs.

* Change device creation logic to use Azure blobs and jobs
* New env var `PCS_AZURE_STORAGE_ACCOUNT` containing the credentials to the Azure storage
* Move custom connection string from file to storage (so that VMs share the custom conn string)

TODO before merge:
* manual verification end-to-end test
* add storage account to ARM template and related env var

TODOs:
* devices created after the simulation won't have the IsSimulated tag
* refactor Devices.cs to allow mocking Azure storage, and add unit tests
* some refactoring needed in the controllers, the complexity of async code is too high, and in one place the code blocks using `.Result`

# Change type 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/255)


Services/Devices.cs
```diff
@@ -357,19 +368,124 @@ public async Task DeleteListAsync(IEnumerable<string> deviceIds)
...
+        {
+            this.instance.InitRequired();
+
+            JobProperties job;
+
+            try
+            {
+                job = await this.registry.GetJobAsync(jobId);
+
+                switch (job.Status)
```
peterfelts: Do we need to care about the default case for this switch?

Services/Devices.cs
```diff
@@ -403,6 +519,67 @@ private int AnalyzeBatchErrors(BulkRegistryOperationResult result)
...
+            }
+
+            return blob;
+        }
+
+        private async Task<CloudBlockBlob> CreateImportExportBlobAsync()
+        {
+            // Container for the files managed by Azure IoT SDK.
+            // Note: use a new container to speed up the operation and avoid old files left over
+            string containerName = ("iothub-" + DateTimeOffset.UtcNow.ToString("yyyy-MM-dd-HH-mm-ss-") + Guid.NewGuid().ToString("N")).ToLowerInvariant();
```
peterfelts: How about adding the IoT Hub name to the containerName? That might help you find the container in Storage Explorer, etc. easier, if we ever need to.

Services/Devices.cs
```diff
@@ -403,6 +519,67 @@ private int AnalyzeBatchErrors(BulkRegistryOperationResult result)
...
+
+            this.log.Info("Creating import blob", () => new { containerName, blobName });
+
+            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(this.config.IoTHubImportStorageAccount);
+            CloudBlobClient client = storageAccount.CreateCloudBlobClient();
+            CloudBlobContainer container = client.GetContainerReference(containerName);
+            CloudBlockBlob blob = container.GetBlockBlobReference(blobName);
+
+            await container.CreateIfNotExistsAsync();
+            await blob.DeleteIfExistsAsync();
```
peterfelts: Why are we deleting the blob right after creating it?

SimulationAgent/SimulationRunner.cs
```diff
@@ -152,146 +147,129 @@ public class SimulationRunner : ISimulationRunner
...
 
-                // Reset rateLimiting counters
-                this.rateLimiting.ResetCounters();
+            foreach (var device in this.devicePropertiesActors)
+            {
+                device.Value.Stop();
             }
+
+            // Allow 3 seconds to complete before stopping the threads
+            Thread.Sleep(3000);
```
peterfelts: Should this be a concurrency config setting, or do we have confidence that three seconds will always be enough time to wait here?

WebService.Test/v1/Controllers/StatusControllerTest.cs
```diff
@@ -58,7 +58,7 @@ public void ItReturnsTheStatusOfRunningSimulation()
             this.SetupSimulationForRunner();
 
             // Act
-            var result = this.target.GetAsync().Result;
+            var result = this.target.GetAsync().CompleteOrTimeout().Result;
```
peterfelts: I like this extension method.

SimulationAgent/SimulationRunner.cs
dluc: no need to make it configurable, and this is going away once the simulation context work is done

PartitioningAgent/Agent.cs
```diff
@@ -83,6 +90,77 @@ public void Stop()
...
+
+            // Check if the device creation is complete
+            if (simulation.DevicesCreationStarted)
+            {
+                this.log.Info("Checking if the device creation is complete...", () => new { SimulationId = simulation.Id });
+
+                var deviceService = this.factory.Resolve<IDevices>();
+                await deviceService.InitAsync(simulation);
+
+                if (await deviceService.IsJobCompleteAsync(simulation.DeviceCreationJobId, () => { creationFailed = true; }))
```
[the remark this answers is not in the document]
dluc: in my opinion the code is self-explaantory, e.g. the boolean is set to true if the device creation failed.  The job completes when the hub platform is done, this is checking if the job is complete.  The return value is a boolean indicating if the job is complete.

Services/Devices.cs
dluc: not needed, all expected cases are handled here, and the method throws an exception if something unexpected is returned

Services/Devices.cs
dluc: it would need extra validation, e.g. I don't know if the chars in the hub name are valid chars in the container name. We can do it separately if finding blobs is too difficult

Services/Devices.cs
dluc: container != blob.  The container is created, and it should be empty.

Services/Devices.cs
brianrei: It would be great if all of our resources followed the same naming convention.  When deploying simulation, many had the same prefix and most had the same suffix but not all.  Can we put this in a shared library or construct the prefix/suffix at the beginning of the deployment?

Services/Devices.cs
brianrei: Why do we need to delete the blob from a new and empty container?

SimulationAgent/Agent.cs
```diff
@@ -75,7 +75,7 @@ public async Task RunAsync()
                     var simulationList = await this.simulations.GetListAsync();
 
                     // currently we support only 1 running simulation so the result should return only 1 item
-                    var runningSimulation = simulationList.FirstOrDefault(s => s.ShouldBeRunning);
+                    var runningSimulation = simulationList.Where(s => s.ShouldBeRunning).FirstOrDefault();
```
hathind-ms: revert this change

PR #256, opened 2018-09-18
Update file type

maple10001:
# Description and Motivation 
* Update file type for uploaded Javascript file.

# Change type 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/256)

### DIFF
--- a/WebService/v1/Controllers/DeviceModelScriptsController.cs
+++ b/WebService/v1/Controllers/DeviceModelScriptsController.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
     [ExceptionsFilter]
     public class DeviceModelScriptsController : Controller
     {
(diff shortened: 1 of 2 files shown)


WebService/v1/Controllers/DeviceModelScriptsController.cs
```diff
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Controller
     [ExceptionsFilter]
     public class DeviceModelScriptsController : Controller
     {
-        private const string APPLICATION_JAVASCRIPT = "application/javascript";
+        private const string APPLICATION_JAVASCRIPT = "text/javascript";
```
hathind-ms: rename const to TEXT_JAVASCRIPT

PR #243, opened 2018-08-29
pfelts/move configurations to storage library

peterfelts:
# Description and Motivation 

In this change, the IoT Hub connection string will be stored and retrieved using the new storage class (StorageRecords), which will talk to DocumentDb directly. The IoT Hub connection string is only stored and retrieved from storage when a custom connection string is used. The use of storage for configuration settings will be needed when multiple simulations are supported. 

# Change type 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Breaking change (breaks backward compatibility)

**Checklist:**

- [x] All tests passed
- [x] The code follows the code style and conventions of this project
- [ ] The change requires a change to the documentation
- [ ] I have updated the documentation accordingly

PR #257, opened 2018-09-19
Fix simulations not starting

peterfelts:
# Description and Motivation 

After my recent pull (#243), simulations weren't starting. Harleen identified the bug: In Simulations.cs, the partitioningComplete flag was getting overwritten with the previous storage value (when simulations were being upserted), so simulations were never being considered ready to start. To fix this I needed a way to know when simulation data models should have their partitioningComplete flag updated and when they should not, so I added an optional flag on UpsertAsync. This flag will only be set when UpsertAsync is called after partitioning is complete. All other calls to UpsertAsync will use the default value for this parameter, which will result in the existing value for the partitioningComplete flag to be used from storage. 

# Change type 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/257)

### DIFF
--- a/Services/Models/Simulation.cs
+++ b/Services/Models/Simulation.cs
@@ -100,8 +100,6 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models
             this.PartitioningComplete = false;
             this.CustomDevices = new List<CustomDeviceRef>();
             this.Statistics = new StatisticsRef();
-
-            this.PartitioningComplete = false;
         }
 
         public class DeviceModelRef
--- a/Services/Simulations.cs
+++ b/Services/Simulations.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         /// <summary>
         /// Create or Replace a simulation.
         /// </summary>
-        Task<Models.Simulation> UpsertAsync(Models.Simulation simulation);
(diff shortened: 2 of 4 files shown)


Services/Models/Simulation.cs
```diff
@@ -100,8 +100,6 @@ public Simulation()
             this.PartitioningComplete = false;
             this.CustomDevices = new List<CustomDeviceRef>();
             this.Statistics = new StatisticsRef();
-
-            this.PartitioningComplete = false;
```
dluc: I would leave the default value explicit, doesn't do any harm

Services/Simulations.cs
```diff
@@ -35,7 +35,7 @@ public interface ISimulations
         /// <summary>
         /// Create or Replace a simulation.
         /// </summary>
-        Task<Models.Simulation> UpsertAsync(Models.Simulation simulation);
+        Task<Models.Simulation> UpsertAsync(Models.Simulation simulation, bool updatePartitioningCompleteFlag = false);
```
dluc: this signature is really odd... upsert should be just upsert, without exceptions. The code used to work in the PoC, what changed?

Services/Models/Simulation.cs
peterfelts: This line was redundant with line 100.

peterfelts:
This was fixed by a different PR: #261

PR #259, opened 2018-09-20
Updates for exposing throttling setting to user for changes

hathind-ms:
# Description and Motivation 
We would use default settings from appsettings if the user doesn't provide these values.
+ Cleaned up some unused references from tests

API schema: 

```json
{
    "ETag": "\"2a008403-0000-0700-0000-5bef24c70000\"",
    "Id": "1760a7df-c555-45e2-b9f7-cb03700b93b8",
    "Name": "Test Simulation - Mixed",
    "Description": "",
    "Enabled": false,
    "DeleteDevicesWhenSimulationEnds": false,
    "Running": false,
    "ActiveNow": false,
    "DevicesDeletionComplete": false,
    "IoTHubs": [
        {
            "ConnectionString": "default",
            "PreprovisionedIoTHubInUse": false,
            "PreprovisionedIoTHubMetricsUrl": null
        }
    ],
    "StartTime": "2018-11-16T15:37:00+00:00",
    "EndTime": "2018-11-30T19:22:48+00:00",
    "StoppedTime": null,
    "DeviceModels": [
        {
            "Id": "truck-01",
            "Count": 5,
            "Override": {
                "Simulation": {
                    "Interval": "00:00:10"
                },
                "Telemetry": [
                    {
                        "Interval": "00:00:10"
                    }
                ]
            }
        }
    ],
    "Statistics": {
        "TotalMessagesSent": 0,
        "AverageMessagesPerSecond": 0,
        "FailedMessagesCount": 0,
        "FailedDeviceConnectionsCount": 0,
        "FailedDeviceTwinUpdatesCount": 0,
        "ActiveDevicesCount": 0
    },
    "RateLimits": {
        "RegistryOperationsPerMinute": 120,
        "TwinReadsPerSecond": 12,
        "TwinWritesPerSecond": 10,
        "ConnectionsPerSecond": 120,
        "DeviceMessagesPerSecond": 120,
        "DeviceMessagesPerDay": 6000000
    },
    "$metadata": {
        "$type": "Simulation;1",
        "$uri": "/v1/simulations/1760a7df-c555-45e2-b9f7-cb03700b93b8",
        "$created": "2018-11-16T20:12:48+00:00",
        "$modified": "2018-11-16T20:12:55+00:00"
    }
}
```

...

# Change type 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/259)


WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            this.ConnectionsPerSecond = 0;
+            this.RegistryOperationsPerMinute = 0;
+            this.TwinReadsPerSecond = 0;
+            this.TwinWritesPerSecond = 0;
+            this.DeviceMessagesPerSecond = 0;
+        }
+
+        // Map API model to service model
+        public static SimulationRateLimits FromServiceModel(Services.Models.Simulation.SimulationRateLimits simulationRateLimits, IRateLimiting rateReporter)
```
peterfelts: Nit: wrap the arguments to each be on a new line.

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            var defaultRatingLimits = rateReporter.GetCounters();
+
+            var connectionsPerSecond = simulationRateLimits?.ConnectionsPerSecond > 0 ? simulationRateLimits.ConnectionsPerSecond : defaultRatingLimits.ConnectionsPerSecond;
```
peterfelts: What happens if simulationRateLimits is null? What does connectionsPerSecond get set to? Perhaps we should have a check at the top of the method for the input not being null?

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Concurrency;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApiModel
+{
+    public class SimulationRateLimits
+    {
+        [JsonProperty(PropertyName = "ConnectionsPerSecond", NullValueHandling = NullValueHandling.Ignore)]
```
peterfelts: Are any of these properties required? If so, we should have a validation method.

WebService.Test/v1/Models/SimulationApiModel/SimulationApiModelTest.cs
```diff
@@ -143,6 +144,20 @@ private void SetupConnectionStringManager()
                 .Returns(Task.CompletedTask);
         }
 
+        private void SetupRateReporter()
+        {
+            this.rateReporter
+                .Setup(x => x.GetCounters())
+                .Returns(new Simulation.SimulationRateLimits
+                {
+                    ConnectionsPerSecond = It.IsAny<int>(),
```
peterfelts: I don't understand this setup. You're specifying htat rateReporter will return a SimulationRateLimits object, but its values are defined is Mock properties... shouldn't we be specifying actual values here for the object that we return?

SimulationAgent/SimulationRunner.cs
```diff
@@ -169,6 +167,9 @@ public void Start(Simulation simulation)
                 // of threads pending on the lock statement
                 this.starting = true;
 
+                // Override throttling limits in case they are provided by user
+                this.rateLimiting.SetCounters(simulation.RateLimits, this.log);
```
peterfelts: Do we want to support changing these limits for a running simulation? If so, we'll need to figure out where these values need to be updated for a running simulation.

Services/Storage/StorageRecord.cs
```diff
@@ -43,7 +43,7 @@ public static StorageRecord FromDocumentDb(Document document)
             var result = new StorageRecord
             {
                 Id = document.Id,
-                ETag = document.GetPropertyValue<string>("ETag"),
+                ETag = document.ETag,
```
peterfelts: I think that I remember that we need line 46 because ETag comes across null, or empty, or something if you read it from the document, and that this will break a test.

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
hathind-ms: If simulationRateLimits  is null, then we will fall back to defaultRatingLimits. I have tested this scenario

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
hathind-ms: These are not required properties

SimulationAgent/SimulationRunner.cs
hathind-ms: For now, these are when we create a new Simulation as they are relative to a IoT Hub throttling limits.

Services/Storage/StorageRecord.cs
hathind-ms: document.GetPropertyValue<string>("ETag") was null.

Services/Concurrency/RatedCounter.cs
```diff
@@ -107,7 +110,7 @@ public long GetPause()
                 this.CleanQueue(now);
 
                 // No pause if the limit hasn't been reached yet,
-                if (this.timestamps.Count < this.eventsPerTimeUnit)
+                if (this.timestamps.Count < this.EventsPerTimeUnit)
```
peterfelts: Nit: I think that our code style uses lower-case for the first character of variable names. I use ReSharper and it tells me that I'm out of compliance when I have variable names start with uppercase characters.

dluc:
given that the API supports an array of hub, what do you thing about storing rate limits on each hub?

Services.Test/SimulationsTest.cs
```diff
@@ -269,7 +269,7 @@ public void UpsertWillFailWhenETagsDoNotMatch()
                 .Wait(Constants.TEST_TIMEOUT);
         }
 
-        [Fact]
+        [Fact(Skip = "This needs updates around ETag"), Trait(Constants.TYPE, Constants.UNIT_TEST)]
```
dluc: OT: we should agree on a (simple) method to track the many TODOs spread across the code base

Services/Concurrency/RateLimiting.cs
```diff
@@ -58,6 +60,50 @@ public class RateLimiting : IRateLimiting
             log.Info("Rate limiting started. This message should appear only once in the logs.");
         }
 
+        public Models.Simulation.SimulationRateLimits GetCounters()
```
dluc: could you change the method name? the method doesn't return counters, does it?

Services/Simulations.cs
```diff
@@ -177,6 +177,7 @@ public async Task<Models.Simulation> InsertAsync(Models.Simulation simulation, s
                 }
             }
 
+
```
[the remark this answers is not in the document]

Services/Concurrency/RateLimiting.cs
hathind-ms: done

Services.Test/SimulationsTest.cs
hathind-ms: Yes, I will think about it. I have fixed this test though

Services/Concurrency/RateLimiting.cs
```diff
@@ -58,6 +60,50 @@ public class RateLimiting : IRateLimiting
...
+            {
+                ConnectionsPerSecond = this.connections.EventsPerTimeUnit,
+                RegistryOperationsPerMinute = this.registryOperations.EventsPerTimeUnit,
+                TwinWritesPerSecond = this.twinReads.EventsPerTimeUnit,
+                TwinReadsPerSecond = this.twinWrites.EventsPerTimeUnit,
+                DeviceMessagesPerSecond = this.messaging.EventsPerTimeUnit
+            };
+        }
+
+        public void SetCounters(
```
dluc: I'd rename this to `SetRateLimits`

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -103,9 +107,9 @@ public Simulation ToServiceModel(string id = "")
                 Enabled = this.Enabled ?? true,
                 StartTime = DateHelper.ParseDateExpression(this.StartTime, now),
                 EndTime = DateHelper.ParseDateExpression(this.EndTime, now),
-                DeviceModels = this.DeviceModels?.Select(x => x.ToServiceModel()).ToList()
+                DeviceModels = this.DeviceModels?.Select(x => x.ToServiceModel()).ToList(),
+                RateLimits = this.RateLimits?.ToServiceModel()
```
dluc: note: when merging in master, this will be slightly different - should be similar to how `PartitioningComplete` is managed, e.g. not allowing overwrite after the simulation has been created

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -165,6 +169,7 @@ public Simulation ToServiceModel(string id = "")
 
             result.DeviceModels = SimulationDeviceModelRef.FromServiceModel(value.DeviceModels);
             result.Statistics = SimulationStatistics.FromServiceModel(value.Statistics);
+            result.RateLimits = SimulationRateLimits.FromServiceModel(value.RateLimits, rateReporter);
```
dluc: `rateReporter` seems a strange var name, could you rename it to match the interface, e.g. `rateLimits` ?

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Jint.Parser.Ast;
```
dluc: this shouldn't be needed, Jint is used only internally and never exposed

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
```diff
@@ -0,0 +1,71 @@
...
+            this.TwinWritesPerSecond = 0;
+            this.DeviceMessagesPerSecond = 0;
+        }
+
+        // Map API model to service model
+        public static SimulationRateLimits FromServiceModel(
+            Services.Models.Simulation.SimulationRateLimits simulationRateLimits,
+            IRateLimiting rateReporter)
+        {
+            var defaultRatingLimits = rateReporter.GetRateLimits();
```
dluc: I would add unit tests to make sure that the defaults are used

Services/Concurrency/RateLimiting.cs
hathind-ms: RateLimits are set in SetCounters. Here we are just reading them.

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
hathind-ms: updated

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
hathind-ms: not sure how this got added. removed now.

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
hathind-ms: added

Services/Concurrency/RateLimiting.cs
```diff
@@ -12,18 +12,20 @@ public interface IRateLimiting
         long GetPauseForNextTwinWrite();
         long GetPauseForNextMessage();
         double GetThroughputForMessages();
+        Models.Simulation.SimulationRateLimits GetRateLimits();
```
avanikp: nit: remove Models.Simulation.

Services/Concurrency/RateLimiting.cs
```diff
@@ -58,6 +60,50 @@ public class RateLimiting : IRateLimiting
             log.Info("Rate limiting started. This message should appear only once in the logs.");
         }
 
+        public Models.Simulation.SimulationRateLimits GetRateLimits()
```
avanikp: same here: remove Models.Simulation.

Services/Models/Simulation.cs
```diff
@@ -125,6 +129,15 @@ public class StatisticsRef
             public double AverageMessagesPerSecond { get; set; }
         }
 
+        public class SimulationRateLimits
```
avanikp: Why not define this model on its own in a separate file ?

Services/Concurrency/RateLimiting.cs
avanikp: Naming seems inconsistent for GetRateLimits and SetCounters. Can we rename these to be consistent ?

Services/Concurrency/RateLimiting.cs
```diff
@@ -58,6 +60,50 @@ public class RateLimiting : IRateLimiting
...
+        {
+            return new Models.Simulation.SimulationRateLimits
+            {
+                ConnectionsPerSecond = this.connections.EventsPerTimeUnit,
```
avanikp: Counters are only set SetCounters for condition > 0, so this.connections can be null if the condition is not true. This will throw an exception here.
Same for other counters.

Services/Concurrency/RateLimiting.cs
```diff
@@ -12,18 +12,20 @@ public interface IRateLimiting
...
     public class RateLimiting : IRateLimiting
     {
         // Use separate objects to reduce internal contentions in the lock statement
 
-        private readonly PerSecondCounter connections;
-        private readonly PerMinuteCounter registryOperations;
-        private readonly PerSecondCounter twinReads;
-        private readonly PerSecondCounter twinWrites;
-        private readonly PerSecondCounter messaging;
+        private PerSecondCounter connections;
```
avanikp: suggest renaming these to connectionsPerSecond, registryOperationsPerMinute, etc for consistency and clarity on the units.

Services/Concurrency/RateLimiting.cs
```diff
@@ -58,6 +60,50 @@ public class RateLimiting : IRateLimiting
...
+            };
+        }
+
+        public void SetCounters(
+            Models.Simulation.SimulationRateLimits simulationRateLimits, 
+            ILogger log)
+        {
+            if (simulationRateLimits.ConnectionsPerSecond > 0)
+            {
+                this.connections = new PerSecondCounter(simulationRateLimits.ConnectionsPerSecond, "Device connections", log);
```
avanikp: How should we handle cases where if condition is false ? Any default values for counters ?

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
```diff
@@ -0,0 +1,71 @@
...
+            this.ConnectionsPerSecond = 0;
+            this.RegistryOperationsPerMinute = 0;
+            this.TwinReadsPerSecond = 0;
+            this.TwinWritesPerSecond = 0;
+            this.DeviceMessagesPerSecond = 0;
+        }
+
+        // Map API model to service model
+        public static SimulationRateLimits FromServiceModel(
+            Services.Models.Simulation.SimulationRateLimits simulationRateLimits,
```
avanikp: nit: remove Services.Models.Simulation.

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
avanikp: suggest renaming GetRateLimits to GetDefaultRateLimits

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -292,5 +302,27 @@ private static SimulationStatistics GetSimulationStatistics(Simulation simulatio
 
             return statistics;
         }
+
+        private IRateLimitingConfig GetRateLimits(IRateLimitingConfig defaultRateLimits)
```
dluc: can you move this into `RateLimitingConfig.ToServiceModel()` ?

hathind-ms: updated

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -70,6 +71,9 @@ public class SimulationApiModel
         [JsonProperty(PropertyName = "Statistics")]
         public SimulationStatistics Statistics { get; set; }
 
+        [JsonProperty(PropertyName = "RateLimits")]
+        public IRateLimitingConfig RateLimits { get; set; }
```
dluc: this is not in line with the other classes and risks to break the API. `IRateLimitingConfig` belongs to the service logic, and could be modified if that changes, but that should not affect the API response schema. There should be a `RateLimits` class in the Web API project which is translated to/from the `RateLimitingConfig` class in the Services project.

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -458,7 +456,49 @@ Content-Type: application/json; charset=utf-8
...
+according to the targeted IoT Hub.
+
+Request:
+```
+POST /v1/simulations
+Content-Type: application/json; charset=utf-8
+```
+```json
+{
+   "ETag":"*",
```
dluc: not needed

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -458,7 +456,49 @@ Content-Type: application/json; charset=utf-8
 }
 ```
 
-## Get details of the simulation, including device models and statistics
+## Creating a simulation with custom rate limits (aka throttling limits)
+
+The following example shows how to override the rate limits for simulation. Service controls
+the outbound requests to IoT Hub in order to avoid getting throttled. These can be changed 
+according to the targeted IoT Hub.
```
dluc: I think the paragraph could be rephrased a bit, I'm not sure about the meaning of the 2nd sentence

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -458,7 +456,49 @@ Content-Type: application/json; charset=utf-8
...
+         "Count":5
+      }
+   ],
+   "RateLimits":{
+      "RegistryOperationsPerMinute":120,
+      "TwinReadsPerSecond":12,
+      "TwinWritesPerSecond":10,
+      "ConnectionsPerSecond":120,
+      "DeviceMessagesPerSecond":120,
+      "DeviceMessagesPerDay":6000000
```
dluc: note that there is no code currently to enforce this setting, it's a TODO

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -458,7 +456,49 @@ Content-Type: application/json; charset=utf-8
...
+   }
+}
+```
+More details about these limits: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling
```
dluc: pls remove the "en-us" from the link so that the user is redirected to the localized version if available

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -458,7 +456,49 @@ Content-Type: application/json; charset=utf-8
...
+
+## Get details of the simulation, including device models, rate limits and statistics
```
dluc: "the simulation" => "a simulation"

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -472,62 +512,75 @@ Content-Type: application/json; charset=utf-8
...
 ```json
 {
-  "ETag": "969ee1fb277640",
-  "Id": "1",
-  "Name": "Sample Simulation",
-  "Description": "This is a sample simulation",
-  "Enabled": false,
-  "Running": false,
-  "IoTHubs": [
+   "ETag":"969ee1fb277640",
```
dluc: the actual ETag has quotes, which could be confusing, so either we leave it out or provide a real example - your call

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -472,62 +512,75 @@ Content-Type: application/json; charset=utf-8
...
+      "$modified":"2018-08-31T00:35:00+00:00"
+   }
 }
 ```
 
 There are some read-only properties in the api response. 
 * `Running` is a calculated property, reporting whether the simulation is
   currently running.
+* `ActiveNow` is a calculated property, reporting whether the simulation is
+  setting up resources in order to start running.
```
dluc: is this description correct?

docs/API_SPECS_SIMULATIONS.md
```diff
@@ -472,62 +512,75 @@ Content-Type: application/json; charset=utf-8
...
+* `DevicesDeletionComplete` is a calculated property, reporting whether the devices has
+  been deleted after simulation has stopped running.
```
dluc: ....after "the" simulation...

WebService/v1/Models/SimulationApiModel/SimulationRateLimits.cs
```diff
@@ -0,0 +1,61 @@
...
+        public int TwinWritesPerSecond { get; set; }
+
+        [JsonProperty(PropertyName = "ConnectionsPerSecond")]
+        public int ConnectionsPerSecond { get; set; }
+
+        [JsonProperty(PropertyName = "DeviceMessagesPerSecond")]
+        public int DeviceMessagesPerSecond { get; set; }
+
+        [JsonProperty(PropertyName = "DeviceMessagesPerDay")]
+        public long DeviceMessagesPerDay { get; set; }
```
dluc: could you add a note saying that this is not implemented yet? i.e. whatever value the user passes, the value is ignored

PR #275, opened 2018-10-09
Devices bulk deletion support

maple10001:
# Description and Motivation 
* Add devices bulk deletion support
* Add `ActiveNow` to simuation api model

# Change type 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

**Todos:**

- [ ] Add tests

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/275)

### DIFF
--- a/PartitioningAgent/Agent.cs
+++ b/PartitioningAgent/Agent.cs
@@ -67,14 +67,23 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.PartitioningAgent
                 if (isMaster)
                 {
                     // Reload all simulations to have fresh status and discover new simulations
-                    IList<Simulation> activeSimulations = (await this.simulations.GetListAsync())
+                    IList<Simulation> simulations = (await this.simulations.GetListAsync());
+
+                    IList<Simulation> activeSimulations = simulations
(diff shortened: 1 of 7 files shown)


Services/Models/Simulation.cs
```diff
@@ -28,6 +28,11 @@ public bool IsActiveNow
         [JsonIgnore]
         public bool DeviceCreationRequired => this.IsActiveNow && !this.DevicesCreationComplete;
 
+        [JsonIgnore]
+        public bool DeviceDeletionRequired => !this.IsActiveNow
```
hathind-ms: Where do we use this property? Do we need to store it in storage?

PartitioningAgent/Agent.cs
```diff
@@ -109,6 +128,63 @@ private async Task CreateDevicesAsync(IList<Simulation> activeSimulations)
             }
         }
 
+        private async Task DeleteIoTHubDevicesAsync(Simulation simulation)
+        {
+            var deletionFailed = false;
+
+            // Check if the device deletion is complete
```
hathind-ms: should the comment be "Check if the device deletion is started"

PR #287, opened 2018-10-16
Stats for simulations spanning multiple nodes

avanikp:
# Description and Motivation
Calculate simulation statistics (TotalMessageCount, TotalFailedMessages, FailedDeviceConnectionsCount, FailedTwinUpdates, Avg msg/sec) for simulation spanning multiple nodes.
Each VM will periodically (30 seconds) write its simulation stats to a common storage collection – one record per simulation per VM. On read, query for all records for a simulation and calculate the aggregate statistics.

Details: 
1) Create a separate collection called “statistics” in cosmosDb (same db where we are other collections – Simulations, Nodes, Partitions, etc). 
2) Each VM (node) will write stats for each simulation once every 30 seconds. There will be one storage record (document) per simulation per node. 
Note: Avg msg/sec will not be stored in db, it will be a calculated prop.

E.g of statistics storage record- 
```
 
"Id":"<SimulationId_nodeId>"
{ 
     "Data":"{
             "SimulationId":"0000-00000-00000", 
             "NodeId":"123245", 
             "Statistics":{"TotalMessagesSent":10000, "TotalFailedMessages":2, "FailedDeviceTwinUpdatesCount":1 ,"FailedDeviceConnectionCount":0
             }"  
} 
```

 
3) Get Simulation API will read all records for a given simulation and calculate aggregate for each stats for all nodes and return aggregated results. 

# Change type 

- [ ] Bug fix
- [X] New feature
- [X] Enhancement

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/287)


Services/Statistics/SimulationStatisticsAgent.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            return statistics;
+        }
+        
+        public async Task CreateOrUpdateAsync(string simulationId, SimulationStatisticsModel statistics)
+        {
+            var nodeId = this.clusterNodes.GetCurrentNodeId();
+            var statisticsRecordId = this.GetStatisticsRecordId(simulationId, nodeId);
+
+            var stats = new SimulationStatisticsRecord
```
hathind-ms: rename to statisticsRecord

Services/Statistics/SimulationStatisticsAgent.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                NodeId = nodeId,
+                SimulationId = simulationId,
+                Statistics = statistics
+            };
+
+            var statsRecord = new StorageRecord
```
hathind-ms: rename to statisticsStorageRecord

Services/Storage/StorageRecords.cs
```diff
@@ -150,6 +151,33 @@ public async Task<IEnumerable<StorageRecord>> GetAllAsync()
...
+
+            try
+            {
+                var query = this.docDb.CreateQuery<DocumentDbRecord>(this.client, this.storageConfig, sqlCondition, parameters).ToList();
+
+                IEnumerable<StorageRecord> storageRecords = query.Select(StorageRecord.FromDocumentDbRecord).ToList();
+
+                // Delete expired records
+                foreach (var record in storageRecords)
+                    if (record.IsExpired())
```
sasathy: foreach () {
      // block of code
}

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -69,8 +72,9 @@ public async Task<SimulationListApiModel> GetAsync()
         public async Task<SimulationApiModel> GetAsync(string id)
         {
             var simulation = await this.simulationsService.GetAsync(id);
+            var stats = await this.simulationStatistics.GetSimulationStatisticsAsync(id);
```
sasathy: don't we need a try catch here?

Services/Statistics/SimulationStatisticsAgent.cs
avanikp: done

Services/Statistics/SimulationStatisticsAgent.cs
avanikp: done

Services/Storage/StorageRecords.cs
avanikp: updated

Services/Simulations.cs
```diff
@@ -284,11 +284,6 @@ public async Task<Models.Simulation> MergeAsync(SimulationPatch patch)
             if (patch.Enabled == false)
```
peterfelts: If the simulation was already stopped, and an PATCH request came in that didn't change Enabled, we'd be updating the StoppedTime here incorrectly. Perhaps this should be:

```suggestion
               if(patch.Enabled == false && item.Enabled == false)
```

Services/Statistics/SimulationStatisticsAgent.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Statistics
+{
+    public interface ISimulationStatistics
+    {
+        Task<SimulationStatisticsModel> GetSimulationStatisticsAsync(string simulationId);
+        Task CreateOrUpdateAsync(string simulationId, SimulationStatisticsModel statistics);
+    }
+
+    public class SimulationStatisticsAgent : ISimulationStatistics
```
peterfelts: I'm not sure that this should be named 'Agent'. This isn't on it's own thread, is it? The other classes that have 'Agent' in their names are on their own threads, looping, interacting with actors. The use of 'Agent' here feels odd to me.

Services/Statistics/SimulationStatisticsAgent.cs
```diff
@@ -0,0 +1,86 @@
...
+            this.log = logger;
+            this.simulationStatisticsStorage = factory.Resolve<IStorageRecords>().Init(config.StatisticsStorage);
+        }
+
+        public async Task<SimulationStatisticsModel> GetSimulationStatisticsAsync(string simulationId)
+        {
+            string sqlCondition = " CONTAINS(ROOT.id, @simulationId)";
+            SqlParameter[] sqlParameters = new[] { new SqlParameter { Name = "@simulationId", Value = simulationId } };
+
+            var simRecords = (await this.simulationStatisticsStorage.GetAsync(sqlCondition, sqlParameters))
```
peterfelts: Calls to GetAsync should probably we wrapped in try/catch.

Services/Statistics/SimulationStatisticsAgent.cs
```diff
@@ -0,0 +1,86 @@
...
+            this.clusterNodes = clusterNodes;
+            this.simulations = simulations;
+            this.config = config;
+            this.log = logger;
+            this.simulationStatisticsStorage = factory.Resolve<IStorageRecords>().Init(config.StatisticsStorage);
+        }
+
+        public async Task<SimulationStatisticsModel> GetSimulationStatisticsAsync(string simulationId)
+        {
+            string sqlCondition = " CONTAINS(ROOT.id, @simulationId)";
```
peterfelts: I don't know if you're validating the input in the caller, but, if not, you should do some input validation here.

Services/Statistics/SimulationStatisticsAgent.cs
```diff
@@ -0,0 +1,86 @@
...
+                Statistics = statistics
+            };
+
+            var statsRecord = new StorageRecord
+            {
+                Id = statisticsRecordId,
+                Data = JsonConvert.SerializeObject(stats)
+            };
+
+            await this.simulationStatisticsStorage.CreateAsync(statsRecord);
```
peterfelts: You'll probably want some try/catch logic here, too, no?

SimulationAgent/SimulationManager.cs
```diff
@@ -93,12 +97,14 @@ public class SimulationManager : ISimulationManager
             IFactory factory,
             IClusteringConfig clusteringConfig,
             ILogger logger,
-            IInstance instance)
+            IInstance instance,
+            ISimulationStatistics simulationStatistics)
```
peterfelts: Does this need to be passed in? Is there a reason that this can't be a singleton for the entire process, or at least just a field of this class?

WebService.Test/v1/Models/SimulationApiModel/SimulationApiModelTest.cs
```diff
@@ -143,6 +143,28 @@ Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApi
...
 
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void ItSetsSimulationStatisticsFromServiceModel()
+        {
+            // Arrange
+            var simulation = this.GetSimulationModel();
+            var statistics = new SimulationStatisticsModel { TotalMessagesSent = 100, FailedDeviceConnectionsCount = 1, FailedDeviceTwinUpdatesCount = 2, FailedMessagesCount = 3  };
+
+            // Act
+            var result = Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApiModel.SimulationApiModel.FromServiceModel(
```
peterfelts: It would be cleaner if you could have a using so that we don't need the entire namespace here.

WebService.Test/v1/Models/SimulationApiModel/SimulationApiModelTest.cs
```diff
@@ -143,6 +143,28 @@ Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApi
...
+                simulation, 
+                statistics);
+
+            // Assert
+            Assert.IsType<Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApiModel.SimulationApiModel>(result);
```
peterfelts: Same here, about the namespace.

Services/Simulations.cs
avanikp: Looks unrelated to the changes in this PR. Could you pls open a separate bug to look into this issue ?

SimulationAgent/SimulationManager.cs
avanikp: This does not need to be a singleton, it is not maintaining any state and is used to read/write data from storage.

WebService/v1/Models/SimulationApiModel/SimulationStatistics.cs
```diff
@@ -43,18 +39,19 @@ public SimulationStatistics()
...
         {
             if (statistics == null) return null;
 
             return new SimulationStatistics
             {
                 TotalMessagesSent = statistics.TotalMessagesSent,
-                AverageMessagesPerSecond = Math.Ceiling(statistics.AverageMessagesPerSecond * 100) / 100
+                FailedDeviceConnectionsCount = statistics.FailedDeviceConnectionsCount,
+                FailedMessagesCount = statistics.FailedMessagesCount,
+                FailedDeviceTwinUpdatesCount = statistics.FailedDeviceTwinUpdatesCount,
```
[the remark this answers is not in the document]
dluc: please let's try not to expose the "twin" concept too much, we're trying not too expose implementation details. I would rename this to "FailedDevicePropertiesUpdates" or similar. Also consider that there are 2 types of updates, device-to-cloud and cloud-to-device, so should we have 2 counters? or does this include both?

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -188,6 +194,15 @@ public static SimulationApiModel FromServiceModel(Simulation value)
             }
```
dluc: looking at lines 166-168 and the conditions at lines 179, 185, 191, I think lines 166-168 are a bug. If StartTime is MinValue, the API response should not contain the field (and similarly for EndTime and StoppedTime)

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -143,7 +147,9 @@ public Simulation ToServiceModel(Simulation existingSimulation, string id = "")
         }
 
         // Map service model to API model
-        public static SimulationApiModel FromServiceModel(Simulation value)
+        public static SimulationApiModel FromServiceModel(
```
dluc: this signature is inconsistent with the usual `FromServiceModel` functions.  What about `SimulationStatisticsModel` being an optional field inside the `Simulation` class instead? See also my comment in controller

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -55,7 +58,10 @@ public async Task<SimulationListApiModel> GetAsync()
         [HttpGet("{id}")]
         public async Task<SimulationApiModel> GetAsync(string id)
         {
-            return SimulationApiModel.FromServiceModel(await this.simulationsService.GetAsync(id));
+            var simulation = await this.simulationsService.GetAsync(id);
+            var stats = await this.simulationStatistics.GetSimulationStatisticsAsync(id);
```
dluc: can we merge these 2 calls, something like:
```
var simulation = await this.simulationService.GetWithStatistics(id)
```

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -188,6 +194,15 @@ public static SimulationApiModel FromServiceModel(Simulation value)
...
 
             result.DeviceModels = SimulationDeviceModelRef.FromServiceModel(value.DeviceModels);
+            result.Statistics = SimulationStatistics.FromServiceModel(statistics);
+
+            if (result.Statistics != null)
+            {
+                DateTimeOffset now = DateTimeOffset.UtcNow;
+                TimeSpan duration = value.IsActiveNow ? now - value.StartTime.Value : value.StoppedTime.Value - value.StartTime.Value;
+                result.Statistics.AverageMessagesPerSecond = result.Statistics.TotalMessagesSent / duration.Seconds;
+            }
```
dluc: this method is getting too long, and adding this code here increases the code complexity. Could you refactor this out.  

Also, make sure the division is not truncating decimals, e.g. `6/10` should return `0.6` and not `0` or `1`.

WebService.Test/v1/Models/SimulationApiModel/SimulationApiModelTest.cs
```diff
@@ -143,6 +143,28 @@ Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApi
                 .Wait(Constants.TEST_TIMEOUT);
         }
 
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
```
dluc: missing test for the average messages per second, making sure the result includes decimals for example

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -188,6 +194,15 @@ public static SimulationApiModel FromServiceModel(Simulation value)
             }
 
             result.DeviceModels = SimulationDeviceModelRef.FromServiceModel(value.DeviceModels);
+            result.Statistics = SimulationStatistics.FromServiceModel(statistics);
+
+            if (result.Statistics != null)
+            {
+                DateTimeOffset now = DateTimeOffset.UtcNow;
+                TimeSpan duration = value.IsActiveNow ? now - value.StartTime.Value : value.StoppedTime.Value - value.StartTime.Value;
```
dluc: is this covered by tests?

SimulationAgent/Agent.cs
```diff
@@ -46,6 +47,9 @@ public class Agent : ISimulationAgent
         // How often (minimum) to log simulation statistics
         private const int STATS_INTERVAL_MSECS = 15000;
 
+        // How often (minimum) to save simulation statistics to storage
+        private const int STATS_INTERVAL_SECS = 30;
```
dluc: please rename these 2 constants, they are too similar.

SimulationAgent/Agent.cs
```diff
@@ -222,6 +227,7 @@ private void StopInactiveSimulations(IList<Simulation> activeSimulations)
 
             foreach (var manager in managersToStop)
             {
+                manager.Value.SaveStatisticsAsync();
```
dluc: is `await` omitted on purpose? if so please add comment explaining the expected behavior when the tasks fail or timeout

SimulationAgent/Agent.cs
```diff
@@ -231,6 +237,23 @@ private void StopInactiveSimulations(IList<Simulation> activeSimulations)
...
+        {
+            foreach (var simulation in simulations)
+            {
+                DateTimeOffset now = DateTimeOffset.UtcNow;
+                TimeSpan duration = now - this.lastPolledTime;
+
+                // Save simulation statistics at specified interval
+                if (duration.Seconds >= STATS_INTERVAL_SECS)
+                {
+                    this.simulationManagers[simulation.Id].SaveStatisticsAsync();
```
dluc: same here, is `await` missing on purpose...etc...

dluc: does this code work? I would expect to throw an exception because the simulation manager is created after invoking this method the first time (see the order of lines 185 and 186)

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
dluc: note that it might take minutes from the StartTime to when the simulation manager starts, e.g. it can take minutes to create devices, so this metric will be quite wrong for the first 10-30 minutes of the simulation. We should consider having a timestamp in the simulation object, tracking when the simulation actually starts, which is after devices and partitions have been created

SimulationAgent/SimulationManager.cs
```diff
@@ -228,6 +234,26 @@ public void PrintStats()
...
+        {
+            try
+            {
+                var simulationModel = new SimulationStatisticsModel
+                {
+                    TotalMessagesSent = this.deviceTelemetryActors.Where(a => a.Key.StartsWith(simulation.Id)).Sum(a => a.Value.TotalMessagesCount),
+                    FailedMessagesCount = this.deviceTelemetryActors.Where(a => a.Key.StartsWith(simulation.Id)).Sum(a => a.Value.FailedMessagesCount),
+                    FailedDeviceConnectionsCount = this.deviceConnectionActors.Where(a => a.Key.StartsWith(simulation.Id)).Sum(a => a.Value.FailedDeviceConnectionsCount),
+                    FailedDeviceTwinUpdatesCount = this.devicePropertiesActors.Where(a => a.Key.StartsWith(simulation.Id)).Sum(a => a.Value.FailedTwinUpdatesCount),
+                };
```
dluc: what happens when the list of actors is empty? e.g. imagine there is a node simulating a single device, and the device is deleted. Will this return 0s or throw an exception?

SimulationAgent/SimulationManager.cs
```diff
@@ -228,6 +234,26 @@ public void PrintStats()
...
         }
 
+        public async Task SaveStatisticsAsync()
+        {
+            try
+            {
+                var simulationModel = new SimulationStatisticsModel
+                {
+                    TotalMessagesSent = this.deviceTelemetryActors.Where(a => a.Key.StartsWith(simulation.Id)).Sum(a => a.Value.TotalMessagesCount),
+                    FailedMessagesCount = this.deviceTelemetryActors.Where(a => a.Key.StartsWith(simulation.Id)).Sum(a => a.Value.FailedMessagesCount),
```
dluc: these 2 lines are looping and filtering the same list twice, please merge into one loop

SimulationAgent.Test/SimulationManagerTest.cs
```diff
@@ -254,6 +258,52 @@ void ItDeletesAllActorsAtTearDown()
...
+                FailedDeviceTwinUpdatesCount = expectedFailedTwinUpdatesCount
+            };
+
+            var mockDeviceConnectionActor = new Mock<IDeviceConnectionActor>();
+            var mockDevicePropertiesActor = new Mock<IDevicePropertiesActor>();
+            var mockDeviceTelemetryActor = new Mock<IDeviceTelemetryActor>();
+
+            for (int i = 0; i < 2; i++)
+            {
+                var deviceName = SIM_ID + $"_{i}";
```
dluc: use one or the other way of merging strings?  either `$"{SIM_ID}_{i}";` or `SIM_ID + "_" + i`

SimulationAgent.Test/SimulationManagerTest.cs
```diff
@@ -254,6 +258,52 @@ void ItDeletesAllActorsAtTearDown()
...
+
+                mockDeviceConnectionActor.Setup(x => x.FailedDeviceConnectionsCount).Returns(3);
+                this.mockDeviceContext.TryAdd(deviceName, mockDeviceConnectionActor.Object);
+
+                mockDevicePropertiesActor.Setup(x => x.FailedTwinUpdatesCount).Returns(5);
+                this.devicePropertiesActors.TryAdd(deviceName, mockDevicePropertiesActor.Object);
+            }
+
+            // Act
+            this.target.SaveStatisticsAsync().Wait();
```
dluc: please use `.CompleteOrTimeout()` instead or `.Wait()`, see other tests in the project

WebService.Test/v1/Controllers/SimulationsControllerTest.cs
```diff
@@ -175,6 +177,27 @@ public void ItReturnsTheSimulationById()
...
+            // Arrange
+            const string ID = "1";
+            var simulation = this.GetSimulationById(ID);
+
+            this.simulationsService
+                .Setup(x => x.GetAsync(ID))
+                .ReturnsAsync(simulation);
+
+            // Act
+            var result = this.target.GetAsync(ID).Result;
```
dluc: use `var result = this.target.GetAsync(ID).CompleteOrTimeout().Result` to make sure the test runs with a limited timeout and doesn't run in the background

Services/Models/SimulationStatisticsModel.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models
+{
+    public class SimulationStatisticsModel
+    {
+        public long TotalMessagesSent { get; set; }
+        public long FailedMessagesCount { get; set; }
+        public long FailedDeviceConnectionsCount { get; set; }
+        public long FailedDeviceTwinUpdatesCount { get; set; }
```
dluc: remove `Count` suffixes

SimulationAgent/Agent.cs
avanikp: no, it was not omitted on purpose.. added await

Services/Statistics/SimulationStatistics.cs
```diff
@@ -0,0 +1,107 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Azure.Documents;
```
dluc: we should not use this dependency here, i.e. the solution should be decoupled from the storage type (in this case Cosmos DB) so that other storage types can be used if needed

Services/Storage/StorageRecords.cs
```diff
@@ -18,6 +18,7 @@ public interface IStorageRecords
     {
         IStorageRecords Init(StorageConfig config);
         Task<StorageRecord> GetAsync(string id);
+        Task<IEnumerable<StorageRecord>> GetAsync(string sqlCondition, SqlParameter[] parameters);
```
dluc: this couples the interface with Cosmos DB. What's the plan if we want to store data in a different storage type, e.g. Cassandra or Azure Tables etc. ?

SimulationAgent/Agent.cs
avanikp: added await.

good catch about the order, the code does work but I had not tried with STATS_INTERVAL_SECS duration too high or low. updated the order.

SimulationAgent.Test/SimulationManagerTest.cs
avanikp: updated

Services/Statistics/SimulationStatistics.cs
```diff
@@ -0,0 +1,107 @@
...
+        public async Task<SimulationStatisticsModel> GetSimulationStatisticsAsync(string simulationId)
+        {
+            if (string.IsNullOrEmpty(simulationId))
+            {
+                this.log.Error("Simulation Id cannot be null or empty");
+                return null;
+            }
+
+            string sqlCondition = " CONTAINS(ROOT.id, @simulationId)";
+            SqlParameter[] sqlParameters = new[] { new SqlParameter { Name = "@simulationId", Value = simulationId } };
```
dluc: this approach couples the solution with Cosmos DB, which is going to be a problem if we want to change storage type, e.g. a customer in RM or partners want to store data differently

WebService.Test/v1/Controllers/SimulationsControllerTest.cs
avanikp: fixed

Services/Statistics/SimulationStatistics.cs
```diff
@@ -0,0 +1,107 @@
...
+            try
+            {
+                var simRecords = (await this.simulationStatisticsStorage.GetAsync(sqlCondition, sqlParameters))
+                    .Select(p => JsonConvert.DeserializeObject<SimulationStatisticsRecord>(p.Data))
+                    .ToList();
+
+                statistics.TotalMessagesSent = simRecords.Sum(a => a.Statistics.TotalMessagesSent);
+                statistics.FailedDeviceConnectionsCount = simRecords.Sum(a => a.Statistics.FailedDeviceConnectionsCount);
+                statistics.FailedDeviceTwinUpdatesCount = simRecords.Sum(a => a.Statistics.FailedDeviceTwinUpdatesCount);
+                statistics.FailedMessagesCount = simRecords.Sum(a => a.Statistics.FailedMessagesCount);
```
dluc: please optimize this code to loop simRecords only once (currently 4 loops)

WebService/v1/Controllers/SimulationsController.cs
avanikp: done

SimulationAgent/SimulationManager.cs
avanikp: added null check for actors

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
avanikp: removed lines 166-168. These values will be set in lines 179, 185, 191

Services/Storage/StorageRecords.cs
avanikp: This class contains all cosmosdb specific requests. I added this so we could specify the query condition.

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
avanikp: added case for avg time.

added ActualStartTime to simulation. This will be set when actors are created for devices in a simulation. For calculating avg, we will use this time as start time.

Services/Statistics/SimulationStatistics.cs
avanikp: Existing implementation (StorageRecords.cs) had cosmosdb specific implementation. I wanted to pass sql query to Get/Delete operation for all simulation stats records. If we want to avoid using SqlParameter and take a dependency on Microsoft.Azure.Documents, I can change query to get/delete one record at a time (i.e call StorageRecord.GetAsync(string id))

Services/Statistics/SimulationStatistics.cs
avanikp: Existing implementation (StorageRecords.cs) had cosmosdb specific implementation. I wanted to pass sql query to Get/Delete operation for all simulation stats records. If we want to avoid using SqlParameter and take a dependency on Microsoft.Azure.Documents, I can change query to get/delete one record at a time (i.e call StorageRecord.GetAsync(string id))

Services/Statistics/SimulationStatisticsAgent.cs
avanikp: thats right, added

Services/Simulations.cs
```diff
@@ -50,6 +54,8 @@ public interface ISimulations
         // Change the simulation, setting the device deletion complete
         Task<bool> TryToSetDeviceDeletionCompleteAsync(string simulationId);
 
+        Task<bool> TryToUpdateSimulation(Models.Simulation simulation);
```
dluc: I don't think this is required, you can use `UpsertAsync`

Services/Simulations.cs
```diff
@@ -414,7 +442,7 @@ public async Task<bool> TryToSetDeviceDeletionCompleteAsync(string simulationId)
             return await this.TryToUpdateSimulation(simulation);
         }
 
-        private async Task<bool> TryToUpdateSimulation(Models.Simulation simulation)
+        public async Task<bool> TryToUpdateSimulation(Models.Simulation simulation)
```
dluc: no need to expose this method, which was created only to support the public one. `UpsertAsync` should be the one to use

dluc: pls add the missing `Async` suffix to the method name

Services/Statistics/SimulationStatistics.cs
dluc: StorageRecords currently supports only DocumentDb but it is designed to support other storage types. This change goes against that design principle. StorageRecords clients shouldn't need to know that DocumentDb is in use. This change will make the port of this code to RM harder, so I encourage to look for an alternative.  You can for example get the full list and filter the records client side, or implement a generic query class designed to support multiple storage engines, like StorageRecords.

Services/Statistics/SimulationStatistics.cs
```diff
@@ -99,6 +102,40 @@ public async Task CreateOrUpdateAsync(string simulationId, SimulationStatisticsM
             }
         }
 
+        public async Task DeleteSimulationStatisticsAsync(string simulationId)
+        {
+            if (string.IsNullOrEmpty(simulationId))
+            {
+                this.log.Error("Simulation Id cannot be null or empty");
```
dluc: I think this should throw an exception, or is it expected that the Id can be null?

SimulationAgent/Agent.cs
```diff
@@ -182,10 +181,10 @@ private async Task RunAsync()
                         .Where(x => x.IsActiveNow).ToList();
                     this.log.Debug("Active simulations loaded", () => new { activeSimulations.Count });
 
-                    await this.SaveSimulationStatisticsAsync(activeSimulations);
                     await this.CreateSimulationManagersAsync(activeSimulations);
+                    await this.SaveSimulationStatisticsAsync(activeSimulations);
                     await this.RunSimulationManagersMaintenanceAsync();
-                    this.StopInactiveSimulations(activeSimulations);
+                    this.StopInactiveSimulationsAsync(activeSimulations);
```
dluc: missing `await`?

SimulationAgent/SimulationManager.cs
```diff
@@ -126,7 +129,7 @@ public class SimulationManager : ISimulationManager
 
             this.simulation = simulation;
             await this.simulationContext.InitAsync(simulation);
-
+
```
dluc: remove spaces

SimulationAgent/SimulationManager.cs
```diff
@@ -466,6 +470,13 @@ private async Task<bool> TryToCreateActorsForPartitionAsync(DevicesPartition par
                     this.CreateActorsForDevice(deviceId, deviceModel, this.deviceCount);
                     this.deviceCount++;
                     count++;
+
+                    // Set ActualStartTime if required
+                    if (!this.simulation.ActualStartTime.Value.Equals(DateTimeOffset.MinValue))
```
dluc: since this is nullable (see also my other comment) you can simplify this expression using `null` and `HasValue`

WebService.Test/v1/Models/SimulationApiModel/SimulationApiModelTest.cs
```diff
@@ -148,21 +147,28 @@ public void ItSetsSimulationStatisticsFromServiceModel()
...
-
+            var statistics = new SimulationStatisticsModel { TotalMessagesSent = 100, FailedDeviceConnections = 1, FailedDevicePropertiesUpdates = 2, FailedMessages = 3  };
+            simulation.Statistics = statistics;
+            var now = DateTimeOffset.UtcNow;
+            simulation.ActualStartTime = now.AddSeconds(-60);
+            simulation.StoppedTime = now;
+            simulation.Enabled = false;
+            // Avg messages = 100/60 (TotalMessagesSent / stoppedTime - startTime)
+            var expectedAvgMessages = 1.6666666666666667;
+
```
dluc: spaces

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -274,5 +263,19 @@ public async Task ValidateInputRequestAsync(ILogger log, IIotHubConnectionString
                 await connectionStringManager.ValidateConnectionStringAsync(iotHub.ConnectionString);
             }
         }
+
+        private static SimulationStatistics getSimulationStatistics(Simulation simulation)
```
dluc: fix method name to start with capital letter

Services/Storage/StorageRecords.cs
dluc: see my last comment, StorageRecords is the only class coupled with Cosmos DB, this couples also the clients of this class

WebService.Test/v1/Models/SimulationApiModel/SimulationApiModelTest.cs
avanikp: fixed

SimulationAgent/SimulationManager.cs
avanikp: updated

Services/Models/Simulation.cs
```diff
@@ -138,6 +147,7 @@ public Simulation()
 
             // By default, run forever
             this.StartTime = DateTimeOffset.MinValue;
+            this.ActualStartTime = DateTimeOffset.MinValue;
```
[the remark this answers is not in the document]
avanikp: updated

Services/Models/Simulation.cs
```diff
@@ -125,6 +126,14 @@ public IList<string> IotHubConnectionStrings
         [JsonProperty(Order = 130)]
         public DateTimeOffset Modified { get; set; }
 
+        // ActualStartTime is the time when Simulation was started
+        [JsonProperty(Order = 140)]
+        public DateTimeOffset? ActualStartTime
+        {
+            get => this.actualStartTime;
+            set => this.actualStartTime = value ?? DateTimeOffset.MinValue;
```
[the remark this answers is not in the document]
avanikp: updated

Services.Test/Statistics/SimulationStatisticsTest.cs
```diff
@@ -18,19 +17,20 @@ namespace Services.Test.Statistics
...
         private readonly Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Statistics.SimulationStatistics target;
         private readonly Mock<IClusterNodes> clusterNodes;
-        private readonly Mock<ISimulations> simulations;
         private readonly Mock<ILogger> log;
         private readonly Mock<IFactory> factory;
         private readonly Mock<IServicesConfig> config;
         private readonly Mock<IStorageRecords> simulationStatisticsStorage;
-        
+        private List<StorageRecord> storageRecords;
+
```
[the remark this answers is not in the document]
avanikp: fixed

SimulationAgent/Agent.cs
avanikp: fixed

SimulationAgent/SimulationManager.cs
avanikp: fixed

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
avanikp: done

Services/Simulations.cs
avanikp: updated

Services/Simulations.cs
avanikp: thanks, updated.

Services/Statistics/SimulationStatistics.cs
avanikp: Id cannot be null, added exception

Services.Test/Statistics/SimulationStatisticsTest.cs
```diff
@@ -26,7 +26,7 @@ public class SimulationStatisticsTest
         private readonly Mock<IServicesConfig> config;
         private readonly Mock<IStorageRecords> simulationStatisticsStorage;
         private List<StorageRecord> storageRecords;
-                
+
```
dluc: spaces

Services/Simulations.cs
```diff
@@ -416,13 +414,13 @@ public async Task<bool> TryToSetDeviceCreationCompleteAsync(string simulationId)
             if (simulation.DevicesCreationComplete) return true;
 
             simulation.DevicesCreationComplete = true;
-
+
```
dluc: spaces

Services/Statistics/SimulationStatistics.cs
```diff
@@ -44,25 +43,26 @@ public async Task<SimulationStatisticsModel> GetSimulationStatisticsAsync(string
...
                     .Select(p => JsonConvert.DeserializeObject<SimulationStatisticsRecord>(p.Data))
                     .ToList();
                  
                 foreach (var record in simulationRecords)
                 {
-                    statistics.TotalMessagesSent += record.Statistics.TotalMessagesSent;
-                    statistics.FailedDeviceConnections += record.Statistics.FailedDeviceConnections;
-                    statistics.FailedDevicePropertiesUpdates += record.Statistics.FailedDevicePropertiesUpdates;
-                    statistics.FailedMessages += record.Statistics.FailedMessages;
+                    if (record.SimulationId == simulationId)
```
dluc: nit: you can apply this filter in line 53

avanikp: sure, done

Services/Simulations.cs
avanikp: fixed

Services.Test/Statistics/SimulationStatisticsTest.cs
avanikp: fixed

WebService/v1/Controllers/SimulationsController.cs
```diff
@@ -28,6 +26,7 @@ public class SimulationsController : Controller
         private readonly IIotHubConnectionStringManager connectionStringManager;
         private readonly IIothubMetrics iothubMetrics;
         private readonly ISimulationAgent simulationAgent;
+
```
dluc: spaces

Services/Simulations.cs
```diff
@@ -297,6 +309,15 @@ public async Task<Models.Simulation> MergeAsync(SimulationPatch patch)
                 }
             }
 
+            if (simulation.Enabled)
+            {
+                // Reset ActualStartTime
+                simulation.ActualStartTime = null;
+
+                // Delete statistics records on simulation start
+                await this.simulationStatistics.DeleteSimulationStatisticsAsync(simulation.Id);
```
dluc: a simulation can be modified also with PUT, so this code won't catch all the cases when a simulation starts. 

There is also a scenario where PUT and PATCH are called multiple times and the app shouldn't delete the metrics every time. For instance PUT and PATCH can be used to modify a running simulation without changing the value of `Enabled`

I think you should move this code in the partitioning agent, and execute it while the partitions and the devices are being created.

avanikp: As we discussed offline, updated PUT to call delete as well. Refactored PATCH so it is clear when the simulation was modified for start/stop.

WebService/v1/Controllers/SimulationsController.cs
avanikp: fixed

WebService.Test/v1/Controllers/SimulationsControllerTest.cs
```diff
@@ -38,7 +36,7 @@ public SimulationsControllerTest()
             this.preprovisionedIotHub = new Mock<IPreprovisionedIotHub>();
             this.simulationAgent = new Mock<ISimulationAgent>();
             this.log = new Mock<ILogger>();
-
+
```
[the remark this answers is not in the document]
avanikp: fixed

Services/Simulations.cs
```diff
@@ -289,16 +294,20 @@ public async Task<Models.Simulation> MergeAsync(SimulationPatch patch)
                 throw new ConflictingResourceException(
                     $"The ETag provided doesn't match the current resource ETag ({simulation.ETag}).");
             }
-
-            if (!patch.Enabled.HasValue || patch.Enabled.Value == simulation.Enabled)
+
```
dluc: spaces

Services.Test/Statistics/SimulationStatisticsTest.cs
```diff
@@ -0,0 +1,162 @@
...
+using Moq;
+using Newtonsoft.Json;
+using Services.Test.helpers;
+using Xunit;
+
+namespace Services.Test.Statistics
+{
+    public class SimulationStatisticsTest
+    {
+        private const string SIM_Id = "1";
```
dluc: fix typo, all upper case

Services.Test/Statistics/SimulationStatisticsTest.cs
```diff
@@ -0,0 +1,162 @@
...
+                FailedDeviceConnections = 6,
+                FailedDevicePropertiesUpdates = 8,
+                FailedMessages = 10
+            };
+
+            SimulationStatisticsRecord expectedStatistics = new SimulationStatisticsRecord
+            {
+                SimulationId = SIM_Id,
+                NodeId = NODE_IDS[0],
+                Statistics= inputStatistics
```
dluc: fix spacing

Services/Models/Simulation.cs
```diff
@@ -142,11 +147,7 @@ public Simulation()
 
             this.DeviceModels = new List<DeviceModelRef>();
             this.CustomDevices = new List<CustomDeviceRef>();
-            this.Statistics = new StatisticsRef();
```
dluc: remove empty line

SimulationAgent/SimulationManager.cs
```diff
@@ -228,6 +237,27 @@ public void PrintStats()
                 });
         }
 
+        public async Task SaveStatisticsAsync()
+        {
+            try
+            {
+                var telemetryActors = this.deviceTelemetryActors ?? this.deviceTelemetryActors.Where(a => a.Key.StartsWith(simulation.Id));
```
dluc: missing `this.` in various places => `this.simulation.Id`

SimulationAgent/SimulationManager.cs
```diff
@@ -440,6 +470,13 @@ private async Task<bool> TryToCreateActorsForPartitionAsync(DevicesPartition par
                     this.CreateActorsForDevice(deviceId, deviceModel, this.deviceCount);
                     this.deviceCount++;
                     count++;
+
+                    // Set ActualStartTime if required
+                    if (!this.simulation.ActualStartTime.HasValue)
+                    {
+                        this.simulation.ActualStartTime = DateTimeOffset.UtcNow;
+                        await this.simulations.UpsertAsync(simulation);
```
dluc: `this.simulation`

WebService.Test/v1/Models/SimulationApiModel/SimulationApiModelTest.cs
```diff
@@ -143,6 +135,35 @@ Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApi
                 .Wait(Constants.TEST_TIMEOUT);
         }
 
+        [Fact, Trait(Constants.TYPE, Constants.UNIT_TEST)]
+        public void ItSetsSimulationStatisticsFromServiceModel()
+        {
+            // Arrange
+            var simulation = this.GetSimulationModel();
+            var statistics = new SimulationStatisticsModel { TotalMessagesSent = 100, FailedDeviceConnections = 1, FailedDevicePropertiesUpdates = 2, FailedMessages = 3  };
```
dluc: fix spacing at the end of the row

WebService.Test/v1/Models/SimulationApiModel/SimulationApiModelTest.cs
```diff
@@ -3,14 +3,6 @@
...
 using System.Collections.Generic;
 using System.Threading.Tasks;
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Concurrency;
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics;
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.IotHub;
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models;
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Runtime;
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent;
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Exceptions;
-using Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService.v1.Models.SimulationApiModel;
```
dluc: after removing this I can compile the code, can you double check?

dluc:
can we remove this from `Services.Models.SimulationPatch` ?

```
public SimulationStatistics Statistics { get; set; }
```

and

```
public class SimulationStatistics
    {
        public long TotalMessagesSent { get; set; }
        public double AverageMessagesPerSecond { get; set; }
    }
```

SimulationAgent/SimulationManager.cs
avanikp: fixed

avanikp:
> can we remove this from `Services.Models.SimulationPatch` ?
> 
> ```
> public SimulationStatistics Statistics { get; set; }
> ```
> and
> 
> ```
> public class SimulationStatistics
>     {
>         public long TotalMessagesSent { get; set; }
>         public double AverageMessagesPerSecond { get; set; }
>     }
> ```

removed

PR #298, opened 2018-10-26
Changes to create default simulation

maple10001:
# Description and Motivation 
* Changes to create default simulation.

# Change type 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/298)


WebService/appsettings.ini
```diff
@@ -19,6 +19,9 @@
 # TCP port where to listen for web service requests. 9003 by default.
 webservice_port = 9003
 
+# Template (name) to be used for solution seeding
+seed_Template = "${?PCS_SEED_TEMPLATE}"
```
dluc: lowercase

WebService/data/device-simulation-template.json
```diff
@@ -0,0 +1,32 @@
+﻿[
```
dluc: 1. I'd move the file under Services/data to avoid confusion
2. the file contains multiple templates, so I'd change the filename suffix to "templates"
3. are these templates just examples? should RM template be here?

SimulationAgent/FileWrapper.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved. 
+
+using System.IO;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent
+{
+    public interface IFile
```
dluc: nite: rename the interface to `IFileSystem`

SimulationAgent/FileWrapper.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved. 
+
```
dluc: psl move this file to Services, it might come handy in other scenarios

SimulationAgent/Agent.cs
```diff
@@ -25,6 +29,7 @@ public interface ISimulationAgent
         Task AddDeviceAsync(string simulationId, string name, string modelId);
         Task DeleteDevicesAsync(List<string> ids);
         void Stop();
+        Task SeedAsync(string templateName);
```
dluc: I don't think having this logic in the SimulationAgent is the right move, it would be great if this agent only responsibility was simulating devices.  What about adding a new agent with this specific responsibility? Looks like many microservices will need a similar feature, so having a defined class hierarchy for this tasks would be reusable across solutions.

WebService/Startup.cs
```diff
@@ -109,6 +109,12 @@ private void StartAgents()
 
             this.simulationAgent = this.ApplicationContainer.Resolve<ISimulationAgent>();
             this.simulationAgent.StartAsync();
+
+            var config = this.ApplicationContainer.Resolve<IConfig>();
+            if (!string.IsNullOrEmpty(config.SeedTemplate))
```
dluc: Seeding should happen only once, only just after the deployment, only in 1 VM. This approach will run the Seed call in every VM and every time the service starts (e.g. after a crash), leading to unexpected situations.

SimulationAgent/FileWrapper.cs
hathind-ms: renamed

SimulationAgent/FileWrapper.cs
hathind-ms: done

WebService/Startup.cs
hathind-ms: Fixed by adding a flag in storage and checking for that before seeding

Services/Simulations.cs
```diff
@@ -367,6 +378,75 @@ public async Task DeleteAsync(string id)
...
+                }
+                else
+                {
+                    this.log.Info("Seeding skipped.");
+                }
+            }
+            catch (Exception e)
+            {
+                var msg = "Failed to seed default simulations." + e.Message;
+                this.log.Error(msg);
```
dluc: include exception

Services/Simulations.cs
```diff
@@ -367,6 +378,75 @@ public async Task DeleteAsync(string id)
...
+            }
+        }
+
+        // This creates sample simulations that will be shown on simulation dashboard by default
+        private async Task SeedSimulationsAsync(string templateName)
+        {
+            string content;
+            var fileName = templateName + ".json";
+            var root = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
+            var filePath = Path.Combine(root, "data", "templates", fileName);
```
dluc: see if constants can make it easier to understand where templates are stored. See also how device models are loaded

Services/Simulations.cs
```diff
@@ -367,6 +378,75 @@ public async Task DeleteAsync(string id)
...
+                    catch (ResourceNotFoundException)
+                    {
+                        simulation.StartTime = DateTimeOffset.UtcNow;
+                        await this.UpsertAsync(simulation);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                this.log.Error("Failed to create sample simulation", ex);
```
dluc: include simulation template details

Services/data/templates/device-simulation-templates.json
```diff
@@ -0,0 +1,32 @@
+﻿[
```
dluc: pleaxe rename the file to indicate this is an example, i.e. a template with multiple simulation

Services/Simulations.cs
```diff
@@ -367,6 +378,75 @@ public async Task DeleteAsync(string id)
...
+                this.log.Debug("Template not found for setting sample simulations.");
+                return;
+            }
+
+            try
+            {
+                var simulationList = JsonConvert.DeserializeObject<List<Models.Simulation>>(content);
+                if (simulationList == null || simulationList.Count == 0) return;
+
+                for (int index = 0; index < simulationList.Count; index++)
```
dluc: please add some docs, explaining that simulation id must be "1" and that we take care of conflicts

WebService/WebService.csproj
```diff
@@ -9,52 +9,61 @@
...
+            <Variable name="PCS_SUBSCRIPTION_DOMAIN" value="" />
+            <Variable name="PCS_SUBSCRIPTION_ID" value="" />
+            <Variable name="PCS_RESOURCE_GROUP" value="" />
+            <Variable name="PCS_IOHUB_NAME" value="" />
+            <Variable name="PCS_SEED_TEMPLATE" value="" />
         </EnvironmentVariables>
         <ApplicationURL>http://localhost:9003</ApplicationURL>
         <LaunchUrl>http://localhost:9003/v1/status</LaunchUrl>
     </PropertyGroup>
+    <ItemGroup>
```
dluc: what is the following block for?

WebService/appsettings.ini
```diff
@@ -19,6 +19,9 @@
 # TCP port where to listen for web service requests. 9003 by default.
 webservice_port = 9003
 
+# Template (name) to be used for solution seeding
```
dluc: add comment about the folder where this is found, and an example, e.g. "foo" => "foo.json"

Services/Simulations.cs

Services/data/templates/multiple-simulations-template.json
```diff
@@ -0,0 +1,32 @@
...
+      {
+        "Id": "truck-01",
+        "Count": 10
+      }
+    ]
+  },
+  {
+    "Name": "Sample Multiple Devices",
+    "Enabled": false,
+    "IotHubConnectionStrings": [
```
dluc: is this correct? looking at the API model I see "IoTHubs"

Services/Simulations.cs
```diff
@@ -367,6 +379,82 @@ public async Task DeleteAsync(string id)
...
 
+        public async Task TrySeedAsync(string templateName)
+        {
+            try
+            {
+                // Skips if the seeding has already been complete
+                if (!await this.mainStorage.ExistsAsync(SEED_STATUS_KEY))
+                {
+                    await this.SeedSimulationsAsync(templateName);
+                    var record = new StorageRecord { Id = SEED_STATUS_KEY, Data = "Seed Completed" };
```
avanikp: Line 390, 391 should be moved inside SeedSimulationsAsync. We should only set the key to "Seed Completed" if the creation successfully completed. In case of an exception (catch block in SeedSimulationsAsync) this should be set to some error state.

Services/Simulations.cs
```diff
@@ -367,6 +379,82 @@ public async Task DeleteAsync(string id)
...
+            string content;
+            var fileName = templateName + ".json";
+            var filePath = Path.Combine(this.config.SeedTemplateFolder, fileName);
+            if (this.fileSystem.Exists(filePath))
+            {
+                content = this.fileSystem.ReadAllText(filePath);
+            }
+            else
+            {
+                this.log.Debug("Template not found for setting sample simulations.");
```
avanikp: What should be SEED_STATUS_KEY in this case ?

Services/Simulations.cs
```diff
@@ -367,6 +379,82 @@ public async Task DeleteAsync(string id)
...
+            Models.Simulation simulation = null;
+
+            try
+            {
+                var simulationList = JsonConvert.DeserializeObject<List<Models.Simulation>>(content);
+                if (simulationList == null || simulationList.Count == 0) return;
+
+                for (int index = 0; index < simulationList.Count; index++)
+                {
+                    // We need to start creating simulations starting with id 1 as it is treated as defailt simulation
```
avanikp: spell check: default

Services/Simulations.cs
```diff
@@ -367,6 +379,82 @@ public async Task DeleteAsync(string id)
             await this.simulationsStorage.DeleteAsync(id);
         }
 
+        public async Task TrySeedAsync(string templateName)
```
dluc: can we move all this new code into Services/Simulations, and from Startup call that?  Having the code here seems quite strange because the agent doesn't use it. I think that Startup making a service call makes it easier to understand what happens

Services/Simulations.cs
```diff
@@ -367,6 +379,82 @@ public async Task DeleteAsync(string id)
...
+                        // create a simulation if no sample simulation exists with provided id.
+                        simulation.StartTime = DateTimeOffset.UtcNow;
+                        await this.UpsertAsync(simulation);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                var message = "Failed to create sample simulation. " + " template: " + JsonConvert.SerializeObject(simulation);
+                this.log.Error(message, ex);
```
avanikp: Need to handle status if SEED_STATUS_KEY in this case

Services/Simulations.cs
```diff
@@ -84,15 +92,19 @@ public class Simulations : ISimulations
...
             IDevices devices,
+            IFileSystem fileSystem,
             ILogger logger,
             IDiagnosticsLogger diagnosticsLogger,
             ISimulationStatistics simulationStatistics)
         {
+            this.config = config;
             this.deviceModels = deviceModels;
             this.storageAdapterClient = storageAdapterClient;
+            this.mainStorage = factory.Resolve<IStorageRecords>().Init(config.MainStorage);
```
dluc: see my other comment: the agent having a direct dependency on the storage is quite strange and inconsistent with the overall design where all the storage calls go via a service (business logic).

Services/Simulations.cs
hathind-ms: this will continue to mark as complete if no or empty template was provided

Services/Simulations.cs
hathind-ms: We don't need to set this key in case of error. We should set it only if this completes successfully.

avanikp: But this will set the key even when SeedSimulationsAsync fails - the exception will be caught in catch block in SeedSimulationsAsync.

Services/Simulations.cs

Services/Simulations.cs
avanikp: Throw from catch block in SeedSimulationsAsync should handle this case.

Services/Simulations.cs
avanikp: Throwing exception from here should handle this case.

Services/Simulations.cs
hathind-ms: resolved other comment

WebService/Startup.cs
```diff
@@ -25,6 +26,9 @@ public class Startup
         // Agent responsible for simulating IoT devices
         private ISimulationAgent simulationAgent;
 
+        // Service responsible for simulating IoT devices
```
dluc: the service is responsible for managing simulations state, not devices

WebService/Startup.cs
```diff
@@ -109,6 +113,14 @@ private void StartAgents()
 
             this.simulationAgent = this.ApplicationContainer.Resolve<ISimulationAgent>();
             this.simulationAgent.StartAsync();
+
+            var config = this.ApplicationContainer.Resolve<IConfig>();
+            if (!string.IsNullOrEmpty(config.SeedTemplate))
```
dluc: nit: this check could be done inside the `TrySeedAsync` method

Services/Simulations.cs
```diff
@@ -71,14 +75,18 @@ public class Simulations : ISimulations
         private const string DEFAULT_SIMULATION_ID = "1";
         private const string DEFAULT_TEMPLATE_NAME = "default";
         private const string DEVICES_COLLECTION = "SimulatedDevices";
+        public const string SEED_STATUS_KEY = "SeedCompleted";
```
dluc: pls fix order of private/public

Services/Simulations.cs
```diff
@@ -387,6 +399,82 @@ public async Task DeleteAsync(string id)
...
+            catch (Exception e)
+            {
+                var msg = "Failed to seed default simulations." + e;
+                this.log.Error(msg);
+                this.diagnosticsLogger.LogServiceError(msg);
+            }
+        }
+
+        // This creates sample simulations that will be shown on simulation dashboard by default
+        private async Task SeedSimulationsAsync(string templateName)
```
dluc: move private methods after public ones

Issue #292, opened 2018-10-23
Update launchSettings.json in WebService to default to system env. variables


avneet723:
# Type of issue 

- [x] Bug
- [ ] New feature
- [ ] Enhancement

# Description 
The help strings in the current launchSettings.json break the default run of device simulation service locally. These strings in the JSON should instead point to the system environment variables in the path and potentially have human readable comments to ask the user to overwrite if desired. 
...

# Steps to reproduce 

1. Open device simulation solution in visual studio. 
2. Build and run the device simulation

# Expected behavior 
Device simulation service starts up and reads the system env. vars by default

# Current behavior 
Delete the line items for the env. variables. This is a bad user experience as it is a required step and out of the box deployment fails. 

# Possible solution 
Point the launchSettings.json to the system env. vars by default like it has been done for other PCS microservices. 

# Context and Environment
* Operating System: Windows
* GitHub branch: master
* .NET Runtime: .NET Core 2.0

# Code to reproduce the bug 
```
using Xunit;

public class MyTest
{
    [Fact]
    public void TestToReproduceIssue()
    {
        // Arrange
        ...

        // Act
        ...

        // Assert
        ...
    }
}
```

PR #299, opened 2018-10-28
Removed the description from the web service connection strings

michal880:
#292
https://github.com/Azure/device-simulation-dotnet/issues/292

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/299)


dluc:
Hi Michal, those descriptions are there on purpose to help the user and there is also some code detecting the free form text to show better error messages, so we don't plan to remove them. The invalid text is also helpful to trigger exceptions with a easy to debug error message in general.

Any reason why you would like to make this change? If so please edit the PR providing some context and rationale, e.g. see the PR template.

Thanks

*Closing the PR for now*

PR #304, opened 2018-10-29
Add ActiveDevices to statistics and other updates

avanikp:
# Description and Motivation 

- Add ActiveDevices to simulation stats (read/write data to stats records).
- Update AvgMessageCount to return value rounded to 2 decimal places.
- Add stats to entire list of simulations in GetAllAsync.
- Update CreateOrUpdateAsync to call Upsert in case the record already exists.
- Add null checks for simulation actual start time and stopped time.
- Add/Update Unit tests and doc.

# Change type 

- [X] Bug fix
- [ ] New feature
- [X] Enhancement

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/304)

### DIFF
--- a/Services/Simulations.cs
+++ b/Services/Simulations.cs
@@ -19,12 +19,14 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
     public interface ISimulations
     {
         // Get list of simulations.
+        // This will have better perf than GetListWithStatisticsAsync.
         Task<IList<Models.Simulation>> GetListAsync();
 
         // Get list of simulations with statistics.
         Task<IList<Models.Simulation>> GetListWithStatisticsAsync();
 
         // Get a simulation.
+        // This will have better perf than GetWithStatisticsAsync.
         Task<Models.Simulation> GetAsync(string id);
 
         // Get a simulation with statistics.
@@ -121,7 +123,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         }
 
         /// <summary>
-        /// // Get list of simulations with statistics.
+        /// Get list of simulations with statistics.
         /// </summary>
         public async Task<IList<Models.Simulation>> GetListWithStatisticsAsync()
         {
--- a/Services/Simulations.cs
+++ b/Services/Simulations.cs
@@ -21,6 +21,9 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         // Get list of simulations.
         Task<IList<Models.Simulation>> GetListAsync();
 
+        // Get list of simulations with statistics.
+        Task<IList<Models.Simulation>> GetListWithStatisticsAsync();
+
         // Get a simulation.
         Task<Models.Simulation> GetAsync(string id);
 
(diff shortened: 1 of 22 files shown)


WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -268,11 +268,22 @@ private static SimulationStatistics GetSimulationStatistics(Simulation simulatio
...
+                {
+                    durationInSeconds = now.Subtract(actualStartTime).TotalSeconds;
+                }
+                else if (simulation.StoppedTime.HasValue)
+                {
+                    durationInSeconds = simulation.StoppedTime.Value.Subtract(actualStartTime).TotalSeconds;
+                }
+
+                statistics.AverageMessagesPerSecond = durationInSeconds > 0 ? Math.Round((double) statistics.TotalMessagesSent / durationInSeconds, 2) : 0;
             }
```
hathind-ms: Use Math. ceil instead:

Math.Ceiling( statistics.TotalMessagesSent / durationInSeconds* 100) / 100

Services/Statistics/SimulationStatistics.cs
```diff
@@ -91,8 +92,17 @@ public async Task CreateOrUpdateAsync(string simulationId, SimulationStatisticsM
 
             try
             {
-                this.log.Debug("Creating statistics record", () => new { statisticsStorageRecord });
-                await this.simulationStatisticsStorage.CreateAsync(statisticsStorageRecord);
+                if (await this.simulationStatisticsStorage.ExistsAsync(statisticsRecordId))
```
peterfelts: I would try to get the existing record, if it exists, and then pass the ETag. If it doesn't exist, you could pass a wildcard for the ETag. The storage library will handle this and will do the right thing with the DocumentDB API.

`var ETag = "*";
var existingRecord = await this.simulationStatisticsStorage.ExistsAsync(statisticsRecordId);
if(existingRecord != null) ETag = existingRecord.ETag;
await this.simulationStatisticsStorage.UpsertAsync(statisticsStorageRecord, ETag);`

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
avanikp: If we want value upto 2 decimal places, using Math.Round should give the expected value. I checked the value returned in both cases.

Services/Statistics/SimulationStatistics.cs
peterfelts: According to the [DocumentDB documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.client.requestoptions?view=azure-dotnet), you can send a meaningless ETag for creation; it'll be ignored. 

"ETag, while valid on Replace* and Delete* operations, would have no impact on a Read*, CreateQuery* or Create* operations."

Services/Simulations.cs
```diff
@@ -21,6 +21,9 @@ public interface ISimulations
         // Get list of simulations.
         Task<IList<Models.Simulation>> GetListAsync();
 
+        // Get list of simulations with statistics.
+        Task<IList<Models.Simulation>> GetListWithStatisticsAsync();
```
peterfelts: Please add a comment explaining that GetList has better performance than calling GetListWithStatisticsAsync.

WebService/v1/Models/SimulationApiModel/SimulationApiModel.cs
```diff
@@ -268,11 +268,22 @@ private static SimulationStatistics GetSimulationStatistics(Simulation simulatio
         {
             var statistics = SimulationStatistics.FromServiceModel(simulation.Statistics);
 
-            if (statistics != null)
+            if (statistics != null && simulation.ActualStartTime.HasValue)
```
peterfelts: Please add a comment explaining why you're checking for simulations that have a value for ActualStartTime. What happens if I pass in a simulation that isn't running?

avanikp: sure added

Services/Simulations.cs
avanikp: added

Services/Statistics/SimulationStatistics.cs
avanikp: Good point about using Upsert for both cases.  In both cases we have similar checks for Exists and pass expected eTag and also it seems cleaner to call Create if the record does not exist. I prefer to keep this implementation.

PR #308, opened 2018-11-02
Replacing deprecated method 'GetDevicesAsync' with 'GetJobsAsync'

sasathy:
Goal:
In order to test read permissions on IoTHub, previously 'getDevicesAsync' method was used which was used to test if the service has read permissions on devices collection in documentDb. But this method has been deprecated. This PR contains the alternate way to test read permissions of IoTHub. 

"GetJobsAsync method of registryManager has been chosen to test read perms. "

This method was chosen because it doesn't affect the quota used by the simulation agent and doesn't interfere with simulations in progress.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/308)


Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -144,7 +147,7 @@ public async Task ValidateConnectionStringAsync(string connectionString)
             if (!match.Groups[CONNSTRING_REGEX_KEY].Value.IsNullOrWhiteSpace())
             {
                 this.ValidateConnectionString(connectionString);
-                await this.TestIoTHubReadPermissionsAsync(connectionString);
+                this.TestIoTHubReadPermissionsAsync(connectionString);
                 await this.TestIoTHubWritePermissionsAsync(connectionString);
```
maple10001: Is there a reason why not await on read `this.TestIoTHubReadPermissionsAsync(connectionString);` but await on write at the next line?
                `await this.TestIoTHubWritePermissionsAsync(connectionString);`

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -172,7 +175,7 @@ private async Task UsePreprovisionedIotHubAsync()
             try
             {
                 this.ValidateConnectionString(this.config.IoTHubConnString);
-                await this.TestIoTHubReadPermissionsAsync(this.config.IoTHubConnString);
+                this.TestIoTHubReadPermissionsAsync(this.config.IoTHubConnString);
```
maple10001: ditto

Services/IotHub/IotHubConnectionStringManager.cs
sasathy: because the function that replaced the deprecated function is no longer asynchronous and so we don't need to wait on it

Services/IotHub/IotHubConnectionStringManager.cs
sasathy: because the function that replaced the deprecated function is no longer asynchronous and so we don't need to wait on it

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -214,13 +217,13 @@ private void ValidateConnectionString(string connectionString)
...
 
-        private async Task TestIoTHubReadPermissionsAsync(string connectionString)
+        private void TestIoTHubReadPermissionsAsync(string connectionString)
         {
             var registryManager = RegistryManager.CreateFromConnectionString(connectionString);
 
             try
             {
-                await registryManager.GetDevicesAsync(1, CancellationToken.None);
+                this.devicesWrapper.GetDevices(registryManager, this.config.DevicesStorage.DocumentDbCollection, this.config.DevicesStorage.DocumentDbPageSize);
```
hathind-ms: why do we need devicesWrapper? this can be replaced by:

registryManager.CreateQuery($"select * from {this.config.DevicesStorage.DocumentDbCollection}", this.config.DevicesStorage.DocumentDbPageSize);

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -144,7 +147,7 @@ public async Task ValidateConnectionStringAsync(string connectionString)
             if (!match.Groups[CONNSTRING_REGEX_KEY].Value.IsNullOrWhiteSpace())
             {
                 this.ValidateConnectionString(connectionString);
-                await this.TestIoTHubReadPermissionsAsync(connectionString);
+                this.TestIoTHubReadPermissionsAsync(connectionString);
```
dluc: missing await

Services/IotHub/DevicesWrapper.cs
```diff
@@ -0,0 +1,19 @@
...
+    public interface IDevicesWrapper
+    {
+        void GetDevices(RegistryManager registryManager, string documentDbCollection, int documentDbPageSize);
+    }
+
+    public class DevicesWrapper : IDevicesWrapper
+    {
+        public void GetDevices(RegistryManager registryManager, string documentDbCollection, int documentDbPageSize)
+        {
+            registryManager.CreateQuery($"select * from {documentDbCollection}", documentDbPageSize);
```
dluc: this code doesn't fetch the devices, it doesn't connect to the hub either

Services/IotHub/IotHubConnectionStringManager.cs
dluc: You should verify that this method passes or fails, depending on the connection string used. As it is this code probably doesn't do anything, and is always passing, did you verify?

dluc: I don't understand why the DocumentDB pagination is used when working with IoT Hub, and given the context of this PR pagination is not really required

WebService/appsettings.ini
```diff
@@ -137,6 +137,8 @@ type = "documentDb"
 documentdb_connstring = "${PCS_STORAGEADAPTER_DOCUMENTDB_CONNSTRING}"
 documentdb_database = "devicesimulation"
 documentdb_collection = "devices"
+# page size for results fetched from document db collection
+documentdb_page_size = 1
```
dluc: to be removed

Services/IotHub/DevicesWrapper.cs
sasathy: replaced this with getJobsAsync function

WebService/appsettings.ini
sasathy: removed

Services/IotHub/IotHubConnectionStringManager.cs
sasathy: CreateQuery was a function that couldn't be waited on. That is why I had to remove the await. Since we are using it no longer, added back the await

Services/IotHub/CheckHubPermissionsWrapper.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved. 
+
+using Microsoft.Azure.Devices;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.IotHub
+{
+    public interface ICheckHubPermissionsWrapper
```
dluc: we use the "Wrapper" suffix when wrapping a class or that doesn't support testing, e.g. some of the SDK classes. I would change the naming here to follow the practice:
1. no need for this new class
2. add a new `GetJobsAsync` method in `RegistryManagerWrapper` - notice how that class is simply a proxy isolating the real code for testability, and the name of the wrapping method is about the same of the original method being called
3. in `IotHubConnectionStringManager.TestIoTHubReadPermissionsAsync()` call `GetJobsAsync` -- please add a comment explaining why, e.g. mentioning that we make a call not affecting the quota used by the simulation agent, to avoid interfering with simulations in progress

sasathy: 1. removed the new class
2. added GetJobsAsync function to RegistryManagerWrapper
3. Added a comment explaining why

Services/IotHub/IotHubConnectionStringManager.cs
sasathy: this has been changed as per our discussion

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -33,17 +33,21 @@ public class IotHubConnectionStringManager : IIotHubConnectionStringManager
         private readonly ILogger log;
         private readonly IDiagnosticsLogger diagnosticsLogger;
         private readonly IStorageRecords mainStorage;
-
+        private readonly IRegistryManager registryManager;
+
```
dluc: please remove spaces

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -33,17 +33,21 @@ public class IotHubConnectionStringManager : IIotHubConnectionStringManager
...
             IDiagnosticsLogger diagnosticsLogger,
+            IRegistryManager registryManager,
             ILogger logger)
         {
             this.config = config;
             this.mainStorage = factory.Resolve<IStorageRecords>().Init(config.MainStorage);
             this.log = logger;
             this.diagnosticsLogger = diagnosticsLogger;
+            this.registryManager = registryManager;
+            this.registryManager.Init(this.config.IoTHubConnString);
```
dluc: what happens if we deploy the solution without a hub and the conn string in the config is empty? what if we have simulations using different conn strings?

Services/IotHub/IotHubConnectionStringManager.cs
```diff
@@ -216,11 +220,11 @@ private void ValidateConnectionString(string connectionString)
 
         private async Task TestIoTHubReadPermissionsAsync(string connectionString)
```
dluc: the code is not testing the conn string passed

sasathy: the connection string itself is tested in a separate call, this.ValidateConnectionString(connectionString).

Services/IotHub/IotHubConnectionStringManager.cs
sasathy: this is the same new line that was present between variables declaration and constructor. I did not add a new empty line

Services/IotHub/IotHubConnectionStringManager.cs
dluc: this method receives in input a connection string to test (`connectionString`), where is is the variable used?

dluc: For instance the two methods `TestIoTHubWritePermissionsAsync` and `TestIoTHubReadPermissionsAsync` which seems similar, work in 2 very different way. Am I missing something?

Services/IotHub/IotHubConnectionStringManager.cs
sasathy: moving initialization of registry manager into the function "TestIoTHubReadPermissionsAsync". This way, the scenarios of connection string being empty or invalid would have been handled even before the function gets called.

Services/IotHub/IotHubConnectionStringManager.cs
sasathy: the connection string passed to TestIoTHubReadPermissionsAsync has the following code in the try block:
                this.registryManager.Init(connectionString);
This is how the connectionString is used.

sasathy: Previously both TestIoTHubReadPermissionsAsync and TestIoTHubWritePermissionsAsync were testing read and write by trying to read devices and add a new device to devices respectively. But the getDevices function used to read devices was deprecated. So, we have to replace that with a different one, which is what we are trying to do through this PR

dluc: I see the code has been added in the last commit, thanks

Services/IotHub/IotHubConnectionStringManager.cs
dluc: there were 8 new spaces in the empty line, I see this has been fixed, thanks

dluc:
Before merging,
1. fix the PR description, currently out of date
2. to be confirmed that "TestReadPermissions" works as expected, running manual tests

PR #313, opened 2018-11-09
Delete all devices even after the simulation is stopped

sasathy:
Currently, user can choose to delete devices associated with a simulation only while creating a new simulation. 

In this PR, we are adding functionality to delete devices associated with the simulation after the simulation has stopped.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/313)


Services/Simulations.cs
```diff
@@ -547,6 +552,11 @@ public async Task<bool> TryToSetDeviceDeletionCompleteAsync(string simulationId)
             simulation.DeviceCreationJobId = null;
             simulation.DevicesCreationStarted = false;
 
+            if (simulation.DeleteDevicesOnce == true)
+            {
+                simulation.DeleteDevicesOnce = false;
```
hathind-ms: don't need the if condition, just set the flag to it's default value i.e. false

sasathy: done

Services/Simulations.cs
```diff
@@ -375,6 +375,11 @@ public async Task<Models.Simulation> MergeAsync(SimulationPatch patch)
             // modify other fields, so we need to check for null
             if (patch.Enabled != null) simulation.Enabled = patch.Enabled.Value;
 
+            if (patch.DeleteDevicesOnce != null)
```
dluc: 'if patch.DeleteDevicesOnce.HasValue`

Services/Simulations.cs
```diff
@@ -546,7 +551,8 @@ public async Task<bool> TryToSetDeviceDeletionCompleteAsync(string simulationId)
             simulation.DevicesCreationComplete = false;
             simulation.DeviceCreationJobId = null;
             simulation.DevicesCreationStarted = false;
-
+            simulation.DeleteDevicesOnce = false;
+
```
dluc: remove spaces

sasathy: I did not add a new space, Devis. This was existing before. In master line 549 will be blank. I am just retaining the exisiting empty line.

Services/Simulations.cs
sasathy: done

Services.Test/Models/SimulationTest.cs
```diff
@@ -151,5 +151,42 @@ public void ItReportsIfDevicesShouldBeCreated()
...
+        [InlineData(false, false, false, false, false)]
+        public void CheckIfDevicesDeletionRequiredIsPopulatedCorrect(
+            bool enabled, 
+            bool deleteDevicesOnce, 
+            bool devicesCreationComplete, 
+            bool deleteDevicesWhenSimulationEnds, 
+            bool expected)
+        {
+            // Arrange and Act
+            var target1 = new SimulationModel
```
hathind-ms: rename var to target

PR #322, opened 2018-11-27
Misc improvements

dluc:
# Description and Motivation 

Note: I suggest to review the PR one commit at a time

**Major:**
1. Backoff in case of d2c messaging throttling, stopping telemetry and twin operations
2. Improve perf reducing load on garbage collector and disposing IoT SDK resources
3. Improve device twin management, e.g. try to avoid twin writes when data is not changed, handle more error scenarios
4. Add "development" feature flag to enable/disable expensive runtime checks

**Minor:**
1. Backoff in case of device count throttling (e.g. when reaching 8000 devices in the free SKU)
2. Remove unused daily counter for telemetry
3. Improve perf reducing the number of no-op async tasks
4. Update IoT SDK and other dependencies
5. Clean up logging code, reimplement log filtering (removed in past PRs)
6. Print SDK version at startup
7. Add some scripts for development, see /scripts/development (create/delete simulation, start storage adapter)

**Bug fixes:**
1. Handle and recover from exceptions in the partitioning agent
2. Fix some swallowed errors/exceptions and unnecessary try/catch
3. Share script interpreter between methods and state to ensure device state consistency
4. Fix logged throughput in case of no traffic, i.e. show 0.0 msg/sec, and round value to 3 decimals
5. Fix logging from ConfigData
6. Change the dev endpoint used to delete simulations: don't delete devices (the endpoint was not working) - no user impact, dev only

# Change type 

- [x] Bug fix
- [x] New feature
- [x] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/322)

### DIFF
--- a/Services/Concurrency/RateLimiting.cs
+++ b/Services/Concurrency/RateLimiting.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Concurrency
             this.messagingQuotaExceeded = false;
             this.messagingQuotaNextProbe = 0;
 
-            // Messaging quota variables
+            // Device quota variables
             this.deviceQuotaLock = new object();
             this.deviceQuotaExceeded = false;
             this.deviceQuotaNextProbe = 0;
--- a/Services/Concurrency/RateLimiting.cs
+++ b/Services/Concurrency/RateLimiting.cs
@@ -74,19 +74,6 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Concurrency
             this.instance.InitComplete();
         }
 
-        public void ChangeClusterSize(int count)
-        {
-            this.log.Info("Updating rating limits to the new cluster size",
-                () => new { previousSize = this.clusterSize, newSize = count });
-
-            this.clusterSize = count;
-            this.connections.ChangeConcurrencyFactor(count);
-            this.registryOperations.ChangeConcurrencyFactor(count);
-            this.twinReads.ChangeConcurrencyFactor(count);
-            this.twinWrites.ChangeConcurrencyFactor(count);
-            this.messaging.ChangeConcurrencyFactor(count);
-        }
-
         public long GetPauseForNextConnection()
         {
             return this.connections.GetPause();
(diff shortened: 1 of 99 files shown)


Services/Diagnostics/Logger.cs
```diff
@@ -14,11 +16,9 @@ public class Logger : ILogger
...
         private readonly string dateFormat;
         private readonly object fileLock;
-
-        public Logger(string processId) :
-            this(processId, new LoggingConfig())
-        {
-        }
+        private readonly ImmutableHashSet<string> blackList;
+        private readonly ImmutableHashSet<string> whiteList;
+        private readonly bool onlyWhiteListed;
```
hathind-ms: Please add a brief comment explaining what 'onlyWhiteListed' means.

Services/Concurrency/RatedCounter.cs
```diff
@@ -182,8 +182,14 @@ public double GetThroughputForMessages()
             {
                 if (this.timestamps.Count > 1)
                 {
+                    var now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+                    var last = this.timestamps.Last();
+
+                    // To avoid stale stats, return 0 Ii there are no events in the last minute
```
hathind-ms: There is a typo in comment  `return 0 if`

SimulationAgent/Agent.cs
```diff
@@ -250,18 +254,10 @@ private async Task SaveSimulationStatisticsAsync(IList<Simulation> simulations)
             {
                 foreach (var simulation in simulations)
                 {
-                    try
+                    if (this.simulationManagers.ContainsKey(simulation.Id))
```
hathind-ms: can simulationManagers be null? 

I think we ran into an issue, with a new deployment, when you create first simulation this was failing. @avanikp - I think you worked on that fix, can you provide more details?

Services/Concurrency/RateLimiting.cs
```diff
@@ -50,6 +86,16 @@ public class RateLimiting : IRateLimiting
             this.log = log;
             this.clusterSize = 1;
             this.instance = instance;
+
+            // Messaging quota variables
+            this.messagingQuotaLock = new object();
+            this.messagingQuotaExceeded = false;
+            this.messagingQuotaNextProbe = 0;
+
+            // Messaging quota variables
```
hathind-ms: Should it be Device quota variables?

SimulationAgent/Agent.cs
dluc: iirc that, or something similar, was a bug I reported to @avanikp during a CR and was fixed. 

The list of managers is a readonly collection initialized in the ctor so it won't be null

Services/Concurrency/RateLimiting.cs
dluc: right, though all of this is going to change given the recent discussions

Services/Diagnostics/Logger.cs
dluc: ok

WebService/DependencyResolution.cs
```diff
@@ -130,10 +130,19 @@ private static void SetupCustomRules(ContainerBuilder builder)
             // dependencies. To improve performance we reuse some instances,
             // for example to reuse IoT Hub connections, as opposed to creating
             // a new connection every time.
+            // Removing these can leads to thousands/millions of new object
```
hathind-ms: nit: `Removing these can lead`

hathind-ms:
Tested with multiple deployments and settings. LGTM

PR #294, opened 2018-10-24
Tagging telemetry messages with user agent 

sasathy:
Currently, the recipients of our telemetry data, ( for example: Hub team) have no way of determining if the message was received from Device Simulation. This could create a lot of problems for them during bursts of incoming traffic. 

And so, in this PR, we are tagging messages with a user agent string - "devicesimulation", indicating the source of the telemetry data.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/294)

PR #326, opened 2018-12-04
Add UserAgent tagging for Remote Monitoring release

jillcary:
# Description and Motivation 

Currently, the recipients of our telemetry data, (for example: IoT Hub team) have no way of determining if the message was received from Device Simulation or Remote Monitoring.

This PR tags messages with a user agent string indicating the source of the telemetry data.

Remote Monitoring is currently using DS-1.0.2, so this PR cherry-picks the commit from https://github.com/Azure/device-simulation-dotnet/pull/294 to the DS-1.0.2 branch.

# Change type 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/326)

### DIFF
--- a/WebService/appsettings.ini
+++ b/WebService/appsettings.ini
@@ -23,6 +23,8 @@ iothub_data_folder = ./data/iothub/
 # and open connections.
 twin_read_write_enabled = "${?PCS_TWIN_READ_WRITE_ENABLED}"
 
+# The user-agent string is used to identify the application to the Azure IoT Hub
+user_agent = "remote_monitoring_device_simulation"
 
 [StorageAdapterService]
 # URL where the storage adapter is available, e.g. http://localhost:9022/v1 on local dev environments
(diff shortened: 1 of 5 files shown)


WebService/appsettings.ini
```diff
@@ -23,6 +23,8 @@ iothub_data_folder = ./data/iothub/
 # and open connections.
 twin_read_write_enabled = "${?PCS_TWIN_READ_WRITE_ENABLED}"
 
+# The user-agent string is used to identify the application to the Azure IoT Hub
+user_agent = "remote_monitoring_device_simulation"
```
timlaverty: How does this get set?  Would someone running DS underneath RM want this string while running it standalone would it be better to be device_simulation_service?  

Should Suite or the CLI set this similar to other env. variables?

hathind-ms: Yes, this value should be set from docker-compose file based on the solution type.

jillcary: I opted to hard-code the string in the appsettings.ini file to a RM specific string because so far as I know only RM uses this release, but you point about someone running simulation standalone is a good point.

I can also update the PR to check against the PCS_SOLUTION_TYPE env var to keep the tag as “devicesimulation” in case it is used by a Device Simulation deployment for some reason.

jillcary:
Opting to upgrade RM to DS-2.0.1 Will submit a new PR to the DS-2.0.1 master branch when ready.

PR #327, opened 2018-12-05
Add support for storing data in Azure Tables

dluc:
# Description and Motivation 

Some data, which doesn't require particular access patterns and rich features of Cosmos DB, can be stored in Azure Tables to reduce the operational cost. This adds the ability to specify whether a collection should be stored in Cosmos DB SQL or Azure Table Storage.

Some implementation details:
* the storage type is set in the main configuration file (appsettings.ini). When using Azure Tables, reference the existing Storage Account connection string, already used to support Azure IoT jobs.
* Azure Tables have a limit of 1MB per record and 64KB per column, so the choice of storing data in Azure Tables must take these limits into account. The code splits data in 12 columns to overcome the limit of 64KB, and throws an error if records are too big to fit.
* all table storage records are stored in the same partition; this is not an issue given the current use cases, but future scenarios might require a different partitioning strategy.

# Change type 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/327)


dluc:
⛑ Work in progress - Do not merge yet ⛑

dluc:
code complete and ready for review

WebService/Startup.cs
```diff
@@ -55,6 +55,7 @@ public Startup(IHostingEnvironment env)
 
             if (ConfigFile.GetDevOnlyConfigFile() != null)
             {
+                Console.WriteLine("===========================\nLOADING SETTINGS FROM " + ConfigFile.GetDevOnlyConfigFile() + "\n===========================");
```
hathind-ms: do we need this console log?

dluc: this is executed only when running the code on a local dev machine, ie not in prod. It is very useful to see this because the custom appsettings path is an env var which could affect solutions in different path. For instance, one might clone the repo many times into:

C:\dev\simulation
C:\workspace\simulation
C:\workspace\simulation_test

but the env var (used by GetDevOnlyConfigFile) is going to be the same, so a dev might be confused without knowing where configuration settings are loaded from.

Services/Storage/TableStorage/DataRecord.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(data),
+                    $"The content length ({len} chars) exceeds the " +
+                    $"maximum record size ({MAX_DATA_LENGTH} chars)");
+            }
+
+            int count = (int) Math.Ceiling((double) len / MAX_PROPERTY_LENGTH);
+            var parts = new string[count + 1];
+            parts[count] = data.Substring(MAX_PROPERTY_LENGTH * (count - 1), len - MAX_PROPERTY_LENGTH * (count - 1));
```
peterfelts: Why are we storing this sub-string in this slot in the 'parts' collection?

PR #331, opened 2018-12-12
Add options to disable device twins and methods

dluc:
# Description and Motivation 

New feature to:
* disable device twins and methods, so one can test scenarios where devices don't establish the extra connections for these hub's features. At the moment the flags are global, and with more work these could become simulation specific settings.
* recreate a device if the device is deleted while a simulation is running

# Change type 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/331)

### DIFF
--- a/Services/DeviceClient.cs
+++ b/Services/DeviceClient.cs
@@ -14,6 +14,7 @@ using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics;
 using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Exceptions;
 using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.IotHub;
 using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Runtime;
 using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Simulation;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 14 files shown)


WebService/appsettings.ini
```diff
@@ -44,17 +37,33 @@ device_models_scripts_folder = ./data/devicemodels/scripts/
...
+# Timeout for the SDK client. By default the SDK uses 4 minutes (240000 msecs).
+sdk_device_client_timeout = 20000
 
 # The user-agent string is used to identify the application to the Azure IoT Hub
 user_agent = "devicesimulation"
 
+# Disable device twin read/writes, to reduce the number of threads and open connections
+# Default: true
+#device_twin_enabled = "${?PCS_TWIN_READ_WRITE_ENABLED}"
+device_twin_enabled = false
```
dluc: ⚡️ this line needs to be replaced by line 58 before merging

WebService/appsettings.ini
```diff
@@ -44,17 +37,33 @@ device_models_scripts_folder = ./data/devicemodels/scripts/
...
+
+# Disable device methods, to reduce the number of threads and open connections
+# Default: true
+#c2d_methods_enabled = "${?PCS_C2D_METHODS_ENABLED}"
+c2d_methods_enabled = false
```
dluc: ⚡️ this line needs to be replaced by line 63 before merging

Services/DeviceClient.cs
```diff
@@ -195,6 +200,12 @@ public async Task SendMessageAsync(string message, DeviceModel.DeviceModelMessag
         /// </summary>
         public async Task UpdatePropertiesAsync(ISmartDictionary properties)
         {
+            if (!this.deviceTwinEnabled)
```
avanikp: q: Do we need to initialize DeviceProperties actor if these global flags are not enabled ? Instead of adding a check here, can we add a check in simulation manager to skip creating device prop actors ?

dluc: this code will never execute because the thread for properties is disabled. I added the code just as an extra protection in case one still wants to have the thread running to stress CPU and memory.

PR #332, opened 2018-12-12
Fixing a regression that prevented properties from being sent

peterfelts:
# Description and Motivation 

@jillcary noticed that device properties weren't being updated in her testing. I looked in to this and saw that device properties were never being sent because the HasWorkToDo method was being overly aggressive in returning false... to the point that the actor would never get in to the state where it would send twin updates. 

# Change type 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/332)


SimulationAgent/DeviceProperties/DevicePropertiesActor.cs
```diff
@@ -172,7 +172,9 @@ public bool HasWorkToDo()
 
             switch (this.status)
             {
+                case ActorStatus.ReadyToStart:
                 case ActorStatus.WaitingForChanges:
+                case ActorStatus.ReadyToUpdate:
```
dluc: pls add also `ReadyToTagDevice`

PR #333, opened 2019-01-08
Replay file  APIs

avanikp:
# Description and Motivation 

APIs to upload, get, validate replay file.

# Change type 

- [ ] Bug fix
- [X] New feature
- [ ] Enhancement

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/333)

### DIFF
--- a/Services/ReplayFileService.cs
+++ b/Services/ReplayFileService.cs
@@ -32,9 +32,9 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         Task DeleteAsync(string id);
 
         /// <summary>
-        /// 
+        /// Validate replay file.
         /// </summary>
-        string Validate(Stream stream);
+        string ValidateFile(Stream stream);
     }
 
     public class ReplayFileService : IReplayFileService
@@ -145,7 +145,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
         /// <summary>
         /// Validate replay file
         /// </summary>
-        public string Validate(Stream stream)
+        public string ValidateFile(Stream stream)
         {
             var reader = new StreamReader(stream);
             var file = reader.ReadToEnd();
@@ -162,8 +162,8 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
                     }
                     catch (MalformedLineException ex)
                     {
-                        this.log.Error("Replay file has inalid csv format", () => new { ex });
-                        throw new InvalidInputException("Replay file has inalid csv format", ex);
+                        this.log.Error("Replay file has inavlid csv format", () => new { ex });
+                        throw new InvalidInputException("Replay file has invalid csv format", ex);
                     }
                 }
             }
--- a/Services/ReplayFileService.cs
+++ b/Services/ReplayFileService.cs
@@ -1,0 +1,174 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Exceptions;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Models;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Runtime;
+using Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Storage;
+using Newtonsoft.Json;
+using Microsoft.VisualBasic.FileIO;
+using FieldType = Microsoft.VisualBasic.FileIO.FieldType;
+
+namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.Services
+{
+    public interface IReplayFileService
+    {
+        /// <summary>
+        /// Get a replay file.
+        /// </summary>
+        Task<DataFile> GetAsync(string id);
+
+        /// <summary>
+        /// Create a replay file.
+        /// </summary>
+        Task<DataFile> InsertAsync(DataFile replayFile);
+
+        /// <summary>
+        /// Delete a replay file.
+        /// </summary>
+        Task DeleteAsync(string id);
+
+        /// <summary>
+        /// 
+        /// </summary>
+        string Validate(Stream stream);
+    }
+
+    public class ReplayFileService : IReplayFileService
+    {
+        private readonly IEngine replayFilesStorage;
+        private readonly ILogger log;
+
+        public ReplayFileService(
+            IServicesConfig config,
+            IEngines engines,
+            IEngine storage,
+            ILogger logger)
+        {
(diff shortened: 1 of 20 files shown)


WebService/v1/Controllers/ReplayFileController.cs
```diff
@@ -0,0 +1,114 @@
...
+                    { StatusCode = (int) HttpStatusCode.BadRequest };
+            }
+
+            return new JsonResult(new ValidationApiModel()) { StatusCode = (int) HttpStatusCode.OK };
+        }
+
+        [HttpPost(Version.PATH + "/[controller]")]
+        public async Task<ReplayFileApiModel> PostAsync(IFormFile file)
+        {
+            if (file == null)
```
peterfelts: I see this same validation done in multiple places. Consider making this a helper?

WebService/v1/Controllers/ReplayFileController.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                this.log.Warn("No data provided");
+                throw new BadRequestException("No data provided.");
+            }
+
+            if (file.ContentType != TEXT_CSV && !file.FileName.EndsWith(".csv"))
+            {
+                this.log.Warn("Wrong content type provided", () => new { file.ContentType });
```
peterfelts: Can we return the expected content type in the error message? If I was calling this API and was told that I was using the wrong content type, I'd appreciate a message telling me what the expected content type is.

Services/ReplayFileService.cs
```diff
@@ -0,0 +1,174 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
peterfelts: How about putting this file in its own folder?

Services/ReplayFileService.cs
```diff
@@ -0,0 +1,174 @@
...
+                parser.SetDelimiters(",");
+                while (!parser.EndOfData)
+                {
+                    try
+                    {
+                        string[] lines = parser.ReadFields();
+                    }
+                    catch (MalformedLineException ex)
+                    {
+                        this.log.Error("Replay file has inalid csv format", () => new { ex });
```
peterfelts: Two typos of 'invalid'

Services/ReplayFileService.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            using (TextFieldParser parser = new TextFieldParser(file))
+            {
+                parser.TextFieldType = FieldType.Delimited;
+                parser.SetDelimiters(",");
+                while (!parser.EndOfData)
+                {
+                    try
+                    {
+                        string[] lines = parser.ReadFields();
```
peterfelts: Why are we reading each line after we've already called reader.ReadToEnd()? Does the file object change at all between when we call ReadToEnd and when we return it?

Services/ReplayFileService.cs
```diff
@@ -0,0 +1,174 @@
...
+                    "Failed to insert new replay file into storage", e);
+            }
+
+            return replayFile;
+        }
+
+        /// <summary>
+        /// Validate replay file
+        /// </summary>
+        public string Validate(Stream stream)
```
peterfelts: It sounds like this method reads the file and does validation. Should this method's name reflect that? It's currently named 'Validate', but that could be considered misleading as it also gets the content.

Services/ReplayFileService.cs
avanikp: fixed

WebService/v1/Controllers/ReplayFileController.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            return ReplayFileApiModel.FromServiceModel(await this.replayFileService.GetAsync(id));
+        }
+
+        [HttpPost(Version.PATH + "/[controller]!validate")]
+        public ActionResult Validate(IFormFile file)
+        {
+            if (file == null)
+            {
+                this.log.Warn("No data provided");
```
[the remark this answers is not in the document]
avanikp: updated

WebService/v1/Controllers/ReplayFileController.cs
```diff
@@ -0,0 +1,114 @@
...
+                throw new BadRequestException("No data provided.");
```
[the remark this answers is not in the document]
avanikp: updated

WebService/v1/Controllers/ReplayFileController.cs
```diff
@@ -0,0 +1,114 @@
...
+            try
+            {
+                var content = this.replayFileService.Validate(file.OpenReadStream());
+            }
+            catch (Exception e)
+            {
+                return new JsonResult(new ValidationApiModel
+                    {
+                        IsValid = false,
+                        Messages = new List<string>
```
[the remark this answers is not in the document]
avanikp: Model expects a list here.

WebService/v1/Controllers/ReplayFileController.cs
avanikp: sure, added.

WebService/v1/Controllers/ReplayFileController.cs
avanikp: sure

Services/ReplayFileService.cs
avanikp: Renamed

Services/ReplayFileService.cs
avanikp: This will read the file and the next block validates each line is in valid format.

WebService/v1/Controllers/ReplayFileController.cs
```diff
@@ -0,0 +1,114 @@
...
+                var content = this.replayFileService.Validate(file.OpenReadStream());
+                replayFile.Content = content;
+                replayFile.Name = file.FileName;
+            }
+            catch (Exception e)
+            {
+                throw new BadRequestException(e.Message);
+            }
+
+            return ReplayFileApiModel.FromServiceModel(await this.replayFileService.InsertAsync(replayFile));
```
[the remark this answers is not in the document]
avanikp: No, this should return the exception

Services/ReplayFileService.cs
avanikp: Since this is the only file, I'll leave it here for now.

Services/ReplayFileService.cs
peterfelts: The new name doesn't really capture that this method returns the file contents. Perhaps call this method 'ValidateAndReadContents'?

Services/ReplayFileService.cs
```diff
@@ -0,0 +1,174 @@
...
+                parser.SetDelimiters(",");
+                while (!parser.EndOfData)
+                {
+                    try
+                    {
+                        string[] lines = parser.ReadFields();
+                    }
+                    catch (MalformedLineException ex)
+                    {
+                        this.log.Error("Replay file has inavlid csv format", () => new { ex });
```
peterfelts: Typo on 'inavlid'.

PR #342, opened 2019-03-05
Custom model send telemetry updates

avanikp:
# Description and Motivation 

1. Advanced device models use long device names (guid.guid.count) which causes storage write exception due to char limit for large number of devices. Update partition size to 2000 to prevent storage error due to char limit.

2. Send telemetry sometimes throws null reference exception in some cases when device state is not set. This repros for custom device model when message size is huge. Added a check in telemetry actor to check for device state actor - if it is in active state then send telemetry. 

# Change type 

- [X] Bug fix
- [ ] New feature
- [ ] Enhancement

**Checklist:**

- [X] All tests passed
- [X] The code follows the code style and conventions of this project

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/342)

### DIFF
--- a/SimulationAgent/DeviceTelemetry/DeviceTelemetryActor.cs
+++ b/SimulationAgent/DeviceTelemetry/DeviceTelemetryActor.cs
@@ -290,7 +290,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent.DeviceTe
 
                 // Looking at the device model, when should the message be sent
                 // note: this.whenToRun contains the time when the last msg was sent
-                var optimalSchedule = this.whenToRun + (long)this.Message.Interval.TotalMilliseconds;
+                var optimalSchedule = this.whenToRun + (long) this.Message.Interval.TotalMilliseconds;
 
                 // TODO: review this approach: when choosing optimalSchedule the app might overload the hub and cause throttling
                 this.whenToRun = Math.Max(optimalSchedule, availableSchedule);
--- a/SimulationAgent/DeviceTelemetry/DeviceTelemetryActor.cs
+++ b/SimulationAgent/DeviceTelemetry/DeviceTelemetryActor.cs
@@ -262,34 +262,58 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.SimulationAgent.DeviceTe
         public void Stop()
         {
             this.status = ActorStatus.Stopped;
+            this.log.Debug("Telemetry stopped",
+                () => new
+                {
+                    this.deviceId,
+                    Status = this.status.ToString()
+                });
         }
 
         private void Reset()
         {
             this.status = ActorStatus.ReadyToStart;
+            this.log.Debug("Telemetry reset",
+                () => new
+                {
+                    this.deviceId,
+                    Status = this.status.ToString()
+                });
         }
 
         private void ScheduleTelemetry()
         {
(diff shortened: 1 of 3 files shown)


WebService/appsettings.ini
```diff
@@ -379,7 +379,7 @@ master_lock_duration = 120000
 # Note that a number bigger than ~10000 can cause errors when storing partitions
 # in Azure Tables due to the records size limit of 1MB.
 # Default: 1000, Min: 1, Max: 10000
-max_partition_size = 5000
+max_partition_size = 2000
```
peterfelts: This was changed from 1000 to 5000 by @dluc in #322. If this size is too small, it looks like we'll have too many storage operations and get throttled. If it's too large, we'll hit the bug that this is attempting to fix. It would be good to understand where the "sweet spot" is for this storage system, given that we'll be writing around this many bytes per device ID:

<guid> + . + <guid> + . + ID

Which, for a 1M simulation would be:

32 + 1 + 32 + 1 + 7 

 or 73 bytes.

For 2000 devices per partition that's 73*2K + some overhead for the meta data that we write, so let's say about 146K. We should, from there, be able to figure out the relationship between when we'll be throttled at different partition sizes.

It would be great if you could figure out what the optimal value is here.

avanikp: Updated to 3K

SimulationAgent/DeviceTelemetry/DeviceTelemetryActor.cs
```diff
@@ -262,36 +262,60 @@ public void HandleEvent(ActorEvents e)
...
+        private void ScheduleTelemetry()
+        {
+            if (this.deviceStateActor.IsDeviceActive)
+            {
+                // Considering the throttling settings, when can the message be sent
+                var availableSchedule = Now + this.simulationContext.RateLimiting.GetPauseForNextMessage();
+
+                // Looking at the device model, when should the message be sent
+                // note: this.whenToRun contains the time when the last msg was sent
+                var optimalSchedule = this.whenToRun + (long)this.Message.Interval.TotalMilliseconds;
```
peterfelts: Please add a space between (long) and _this_

SimulationAgent/DeviceTelemetry/SendTelemetry.cs
```diff
@@ -35,30 +35,40 @@ public void Init(IDeviceTelemetryActor context, string deviceId, DeviceModel dev
...
-            this.log.Debug("Checking to see if device is online", () => new { this.deviceId });
-            if (!state.ContainsKey("online") || (bool) state["online"])
+            try
             {
-                this.log.Debug("The device state says the device is online, sending telemetry...", () => new { this.deviceId });
+                var state = this.context.DeviceState.GetAll();
+
+                // device could be rebooting, updating firmware, etc.
+                this.log.Debug("Checking to see if device is online", () => new { this.deviceId });
+                if (!state.ContainsKey("online") || (bool)state["online"])
```
peterfelts: Formatting (spaces)

avanikp: This is existing code. Added a space though

SimulationAgent/DeviceTelemetry/DeviceTelemetryActor.cs
avanikp: added

PR #344, opened 2019-03-06
Update dot net version

avanikp:
# Description and Motivation 

Update to dotnet version 2.1

# Change type 

- [ ] Bug fix
- [ ] New feature
- [X] Enhancement

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/device-simulation-dotnet/344)

### DIFF
--- a/WebService/Startup.cs
+++ b/WebService/Startup.cs
@@ -18,6 +18,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using ILogger = Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics.ILogger;
+using Microsoft.AspNetCore.Mvc;
 
 namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService
 {
@@ -71,7 +72,7 @@ namespace Microsoft.Azure.IoTSolutions.DeviceSimulation.WebService
             services.AddCors();
 
             // Add controllers as services so they'll be resolved.
-            services.AddMvc().AddControllersAsServices();
+            services.AddMvc().AddControllersAsServices().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
 
             // Prepare DI container
             this.ApplicationContainer = DependencyResolution.Init(services);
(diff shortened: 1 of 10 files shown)


WebService/Startup.cs
```diff
@@ -18,6 +18,7 @@
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using ILogger = Microsoft.Azure.IoTSolutions.DeviceSimulation.Services.Diagnostics.ILogger;
+using Microsoft.AspNetCore.Mvc;
```
peterfelts: Do we know why this is needed?

avanikp: This is needed for the changes in line 75 (as per documentation for net core 2.1 upgrade) -

services.AddMvc().AddControllersAsServices().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);

scripts/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM microsoft/dotnet:2.0.3-runtime-jessie
+FROM microsoft/dotnet:2.1-aspnetcore-runtime
```
saixiaohui: Please change the maintainer to our team's alias

WebService/Startup.cs
```diff
@@ -71,7 +72,7 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
             services.AddCors();
 
             // Add controllers as services so they'll be resolved.
-            services.AddMvc().AddControllersAsServices();
+            services.AddMvc().AddControllersAsServices().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
```
saixiaohui: do we need this additional function call? what's the downside of not having it? what's the benefit for having it?

scripts/docker/Dockerfile
avanikp: updated

WebService/Startup.cs
avanikp: I'm adding it as per documentation here to get the latest behavior: https://docs.microsoft.com/en-us/aspnet/core/mvc/compatibility-version?view=aspnetcore-2.1

Else the default will set compatibility version to 2.0

PR #364, opened 2019-10-25
Fixing the elevator model

troyhopwood:
# Description and Motivation 

...

# Change type 

- [ ] Bug fix
- [ ] New feature
- [ ] Enhancement

**Checklist:**

- [ ] All tests passed

### DIFF
--- a/Services/data/devicemodels/elevator-01.json
+++ b/Services/data/devicemodels/elevator-01.json
@@ -12,7 +12,8 @@
             "vibration": 10.0,
             "vibration_unit": "mm",
             "temperature": 75.0,
-            "temperature_unit": "F"
+            "temperature_unit": "F",
+			"moving": true
         },
         "Interval": "00:00:10",
         "Scripts": [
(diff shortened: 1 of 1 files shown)


Services/data/devicemodels/elevator-01.json
```diff
@@ -12,7 +12,8 @@
             "vibration": 10.0,
             "vibration_unit": "mm",
             "temperature": 75.0,
-            "temperature_unit": "F"
+            "temperature_unit": "F",
+	    "moving": true
```
peterfelts: This change makes me think that we don't even use/consume the default state in the JS files [here](https://github.com/Azure/device-simulation-dotnet/blob/1ada7f41ea8856f9a00d0be455571b87bcbc1ea0/Services/data/devicemodels/scripts/elevator-01-state.js#L20) and [here](https://github.com/Azure/device-simulation-dotnet/blob/1ada7f41ea8856f9a00d0be455571b87bcbc1ea0/Services/data/devicemodels/scripts/elevator-02-state.js#L20). 

It would be great if this change included a comment in all the JS files that called out that we don't respect these default values if they're provided in the device model (I assume that is why you added the value here).

